Issue #25, opened 2015-01-26
Ignore symlinks


LeeXGreen:
Hi,

I've been running RubyCritic on a large Rails project. We use Rails engines to organize / separate our code, and for better or worse, we have a number of symlinked files from one engine to another or back to the main app. (I'd like to remove these links eventually but that's another discussion.)

When we run RubyCritic on this project, we get a lot of duplicate code warnings, and the source/target of the duplication are the source/target of the symlinks. It would be nice if RubyCritic could be configured to ignore symlinks.

I know that my use case might not be very common, so I'm prepared to work on this myself. I just wanted to open an issue here to answer a few questions:

* is this something that should be implemented at the RubyCritic level?
  * I think so, given the SourceLocator class
* should RubyCritic just ignore symlinks by default?
  * if not, should this be a CLI option?

Note -- this is RubyCritic version 1.1.1 running under JRuby 1.7.9 (which implements Ruby 1.9.3)

guilhermesimoes:
I am sensible of your problem and your proposed solution looks good to me.

Regarding your questions:

> is this something that should be implemented at the RubyCritic level?
> * I think so, given the SourceLocator class

I think so too. But I'm surprised this issue hasn't come up before in more well known gems such as RuboCop or Reek.

> should RubyCritic just ignore symlinks by default?
> * if not, should this be a CLI option?

I'm torn on this. I think it definitely should ignore a symlink if it points to a file that will be also included. However, if it points to a file outside of the current directory tree then it shouldn't be ignored.

For now, making this a CLI option (turned off by default) is a good compromise and (most importantly) backwards compatible.

Also, good to know that RubyCritic is running well under JRuby :smile:

guilhermesimoes:
Closed by #26 :clap:

PR #26, opened 2015-01-26
Optionally ignore symlinks with -i

LeeXGreen:
Here's a first draft at what the functionality I outlined in #25.


LeeXGreen:
I see why the Travis build failed, even though it passed on my dev machine.

I've guaranteed that only one of [original file, symlink] is processed by RubyCritic, but there's no guarantee that the original file is the one that remains. Whichever one is seen first will most likely be the winner.

LeeXGreen:
Okay. This will now prefer files over symlinks, unless there are no real files for a given realpath. (This is a bit of a pathological case but it is possible if you have files completely external to your repo that are linked into your repo.)

This is failing a rubocop style check due to the length of https://github.com/LeeXGreen/rubycritic/blob/option_ignore_symlinks/lib/rubycritic/cli/options.rb#L11-L36 but I don't really see a sensible way to make this shorter.

guilhermesimoes:
Looks good to me! Still, I'd like to take a better look at this by the end of the week. I know the changeset is small but bear with me.

And don't worry too much about the RuboCop checks failing. I set up RuboCop not long ago so its defaults may be a bit far from sane.

LeeXGreen:
Sure thing -- and now that I've had a chance to sleep on this code, I think the command-line option could be better named to make it clear that it does not so much *ignore* symlinks, as *de-duplicate* them.

I especially welcome your input here.

LeeXGreen:
Just found out this will not work properly on JRuby thanks to https://github.com/jruby/jruby/issues/538.

It should still be achievable, though -- just might require special-case behavior for JRuby.

LeeXGreen:
Okay, this is behaviorally correct on JRuby now :)

guilhermesimoes:
I knew I'd forgotten something last week.

Was just checking out the code when I noticed you changed this just now to work with JRuby. Here's the thing though, the previous tests were already passing on JRuby, so either it was already working or we're missing a test.

LeeXGreen:
Aha -- it's a JRuby 1.7.9 vs 1.7.18 difference.

JRuby 1.7.9 needs the fixes in 5f3391c, whereas 1.7.18 was good with only 7447929.

Since the RubyCritic TravisCI build only runs 1.7.18, we passed with the first build.

LeeXGreen:
To be precise, this is fixed for Jruby >= 1.7.14

guilhermesimoes:
I see, you like to live on the edge!

A couple points before I merge though:

* In the JRuby issue you linked to, there's a reference to [this other commit][1]. And I was just wondering if there's no need to `require "java"` like it's being done there.

* Also, if you can, rename the option prefix `ignore_` to `deduplicate_`.

* And remove the shorthand option `"-i"`, just make it an empty string.

Like I said in the original issue, deduplicating paths makes too much sense to be optional but I guess we can make this a default later on.

And sorry for taking so long to check this.

[1]: https://github.com/rapid7/metasploit_data_models/commit/0b4fc033ea45309813e61446db8f10cbb18b75a1

LeeXGreen:
Okay, your points from above are fixed:

* Not sure whether we need to `require "java"`, so I threw it in.
* it's now `deduplicate_` throughout
* `-i` option now removed.

Here is a clean commit if you prefer this one:

https://github.com/LeeXGreen/rubycritic/commit/78669f28d46057e10d842b654e4cec24e058ffa9

guilhermesimoes:
Done! It's now on master.

Yesterday I started working on a JSON formatter which I want in the next release. So a new version should be released sometime this week when that's done.

Once again sorry for taking so long with this. Thanks! :smile:

PR #29, opened 2015-02-19
Suppress ratings.

halostatue:
Rubycritic provides useful reports, but I’m completely uninterested in having letter grades assigned to code. It is the main reason that I will not use Code Climate.

I can understand if you’re not interested in this particular pull request, but I think that this is a *good* change for more mature teams.

I don’t think it would be bad to offer a Rubycritic configuration file for projects to use, possibly in conjunction with guard-rubycritic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 1.3.0 / 2015-02-16
+
+* [FEATURE] New CLI option `--deduplicate-symlinks` to deduplicate symlinks (by LeeXGreen)
+* [CHANGE] Update to Reek 1.6.5 (from 1.6.3)
+* [CHANGE] Remove ruby2ruby dependency
+
 # 1.2.1 / 2015-01-17
 
 * [FEATURE] Support Ruby 2.2
@@ -8,7 +14,7 @@
 
 * [FEATURE] Add CI mode that only analyses the last commit
 * [FEATURE] Add partial support for Mercurial
-* [FEATURE] Allow using Rubycritic programatically
+* [FEATURE] Allow using RubyCritic programatically
 * [CHANGE] Update to Reek 1.6.0 (from 1.3.8)
 
 # 1.1.1 / 2014-07-29
--- a/lib/rubycritic/cli/options.rb
+++ b/lib/rubycritic/cli/options.rb
@@ -25,6 +25,10 @@ module Rubycritic
             @deduplicate_symlinks = true
           end
 
+          opts.on("--suppress-ratings", "Suppress letter grade ratings") do
+            @suppress_ratings = true
+          end
+
           opts.on_tail("-v", "--version", "Show gem's version") do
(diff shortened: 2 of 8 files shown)


guilhermesimoes:
Hm, did you forget to add other changes to your pull request or am I supposed to finish what you started? Just asking :smile: 

Regarding the change in question. I think grades are tremendously helpful in prioritizing the code that should be fixed. I think the problem lies in us humans who always strive to achieve the highest scores, get all A's and increase our karma counts, whether on Reddit or on StackOverflow.

Even the guys at Code Climate struggle to get across the message that scores shouldn't be taken at face value. Bryan Helmkamp, one of the creators of Code Climate, created [this secret gist][1] one year ago to expose his thoughts on the issue. The first paragraph is succinct:

> Sometimes, when people use Code Climate, they try to make changes in response to all of the issues it reports, in order to achieve straight A's and/or a 4.0 GPA. We do not recommend using Code Climate in this way.

I completely agree with Bryan and I believe RubyCritic should be used in the same vein.

Now, I'm not completely opposed to this proposal but I'd prefer if we worked towards grades that:

1. Are more self-explanatory, and
2. Better reflect the *maintainability* of code

I'm not sure if it's noticeable but Reek smells don't even affect grades. Only Flog's complexity and Flay's duplication are taken into account when assigning grades. At one point I even considered ripping out grades altogether because I didn't think the scoring formula was good enough (I still think it could be improved). But now there are people who wish RubyCritic provided a project wide grade...

I think your team is making a mistake by refusing to use Code Climate because of its letter grades (price, now that I would understand).

But anyway, like I said, I'm not completely opposed to suppressing grades. I just wish they were so good that you wouldn't want to suppress them.

[1]: https://gist.github.com/brynary/21369b5892525e1bd102

halostatue:
@guilhermesimoes Oh. Oops. Let me get that fixed—I had made the changes and then I think I managed to lose the template changes that I had made.

I personally don’t care what the *intent* of the ratings/grades are, they are a mistake across the board for all of the negative reasons you outlined. More importantly, I never want these “grades” visible to anyone outside of the development team itself, because all of the subtleties that Bryan & you talk about go out the window when you leave the development team, or add new team members…it turns into a problem.

IMO, rubycritic provides a nicer view than metric_fu, and that is its real benefit. I’ll always want to suppress computed ratings because I don’t believe them to be useful and believe them to be actively harmful in all cases.

guilhermesimoes:
> I never want these “grades” visible to anyone outside of the development team itself, because all of the subtleties that Bryan & you talk about go out the window when you leave the development team, or add new team members

That's a very good point, and something I hadn't even considered. I think you made your case and I am happy that RubyCritic is able to accommodate your team's needs. Your code is now merged (in 79ffec3) :grinning:

And not to gloat or anything, but I also think RubyCritic's view is better than metric_fu's. I think it could look even better though!

Issue #9, opened 2014-07-23
JSON template to help analisys


stupied4ever:
I think that a json template would help to generate post build jenkins plugins to send emails on class degradation and other things what do you think? I could work on that, what do you think?

suryagaddipati:
Great gem!!

I was wondering if another generator(json?) was ever built. The only option right now seem to be html reporter.

guilhermesimoes:
Nope, nothing so far I believe. Though it shouldn't be *too* hard. I can probably tackle this on the next weekend.

ancorgs:
I have already tried the new json formater and it rocks. Just a suggestion: could you include the cost for each analysed_module in addition to the rating? Letters are great for a quick review by a human eye, but when you are exporting to json you most likely want to process the output and in that contexts, numbers are more useful (and precise).

On the other hand, in https://github.com/whitesmith/rubycritic/wiki/Roadmap says that a project rating is planned. I think I could contribute that but I wonder what was exactly the plan to compute it. Just the average of the cost of all analyzed modules or something more complex?

ancorgs:
It was actually completely trivial, so I created #31

guilhermesimoes:
Regarding the project rating: that's precisely the thing. How do we compute it? I feel like if the project rating is wrongly implemented it may either:

1. Not be valuable since it will always be too high.

2. Upset people since it will always be too low.

Also, from the moment we implement the formula and give a project rating, people will expect the rating to remain constant. So if we want to tweak the formula, how do we make sure that we're not setting off alarms half-way around the world? Ideally we'd have a test suite for this.

For such a "simple" feature, it sure seems complicated to me. But I'm open to ideas! I think it would be great if you opened a pull request where we could have a more in-depth discussion :grinning:

guilhermesimoes:
Btw, @ancorgs, since you've been playing with the new JSON formatter... I just checked Reek's and Rubocop's JSON formatters and they both output the raw JSON to the console instead of creating a file. Do you think RubyCritic's should behave that way as well?

ancorgs:
Outputting to standard output is actually more flexible. In my case, I'm interested in the JSON formatter because I want to process the output, so I would actually prefer to have it on the standard output instead of a file.

guilhermesimoes:
@ancorgs Thanks for the feedback!

@suryagaddipati Took me a big longer than I expected but the JSON formatter is done. Try the new version 1.4.0!

For improvements on the JSON formatter, open a new issue :grinning:

PR #31, opened 2015-03-04
Add cost to analysed_modules in the JSON report

ancorgs:
As suggested in https://github.com/whitesmith/rubycritic/issues/9#issuecomment-77187725


suryagaddipati:
:+1:

guilhermesimoes:
Wow, thanks for spotting this. That JSON formatter is kinda a WIP so I will just amend this change as well as a bunch of other things that are still not quite right (I like to have all related things in the same commit).

Thanks!

PR #32, opened 2015-03-04
Add GPA calculations

natesholland:
I would like to add the ability to calculate an over all GPA to rubycritic and I need some feedback. What I have right now is a rough sketch and needs to be cleaned up and advised upon. Before this is done I would like to add the ability to fail if the GPA is below a determined threshold. It would also be cool to be able to fail if an F shows up. Also I will clean up my commits before they get merged in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 1.4.0 / 2015-03-14
+
+* [FEATURE] New report in JSON format. Available by using the new CLI option `-f`
+* [FEATURE] New CLI option `--suppress-ratings` to suppress ratings (by halostatue)
+* [CHANGE] Improve UI, particularly the sortable tables (by crackofdusk)
+
 # 1.3.0 / 2015-02-16
 
 * [FEATURE] New CLI option `--deduplicate-symlinks` to deduplicate symlinks (by LeeXGreen)
@@ -16,6 +22,7 @@
 * [FEATURE] Add partial support for Mercurial
 * [FEATURE] Allow using RubyCritic programatically
 * [CHANGE] Update to Reek 1.6.0 (from 1.3.8)
+* [BUGFIX] Fix issue #18 - Prevent encoding issues when using Git
 
 # 1.1.1 / 2014-07-29
 
--- a/lib/rubycritic/version.rb
+++ b/lib/rubycritic/version.rb
@@ -1,3 +1,3 @@
 module Rubycritic
-  VERSION = "1.3.0"
+  VERSION = "1.4.0"
 end


natesholland:
One thing I would like feedback from others on is, should the GPA for each file be weighted equally or should they be weighted by total number of lines?

guilhermesimoes:
I'm sorry for doing this to you but I force pushed to master to clean up some stuff. I promise I won't repeat this feat. You can `rebase -i master` and delete the first four commits to solve most of the problems but there will be some conflicts as the file `lib/rubycritic/reporters/base.rb` no longer exists. You should put the code that you placed there into its own class (like GlobalRatingCalculator) and for now require it in each `_report` file.

guilhermesimoes:
As for weighting the ratings by total number of lines, it makes total sense to me! I'd like to hear other opinions though.

natesholland:
No worries about the force push, I will resolve that. I polled some other guys at my office and they also thought it should be weighted by line.

natesholland:
Ok I just fixed all the merge requests, I'll go back later and read over the aftermath of that and see if my designs are still ok or if they need to be changed. I still have some rubocop things that are failing but otherwise this should work. I added the ability to exit with non-zero status codes if the over all gpa falls below a certain level, or if an F (or D or C or B) appears. I know at Spiceworks we would like to use this with our CI servers. I'll clean up some of the rubocop stuff soon and then look into writing my own tests for this stuff.

natesholland:
Ok I've fixed as much of the rubocop stuff as I can but ultimately I don't know if it's possible to parse all the options in one method and keep the complexity low enough. I think at this point I just need someone else to read over the change set and give me their thoughts. @LeeXGreen could you take a look at this now?

LeeXGreen:
Well, we can satisfy RubyCritic by using this idiom in `lib/rubycritic/configuration.rb`:

```diff
-      self.mode = options[:mode] || :default
+      self.mode = options.fetch(:mode) { :default }
```

Or we could define defaults for each in a separate method/constant, then merge or reverse_merge those with the options hash before setting them on the Configuration object.

With respect to the CLI `parse` method -- we could probably think of ways to split this up into stanzas, but I'm not sure if that's more readable, or *less* readable.



Other than the Rubocop issue, there is maybe some more DRYing up to be done of the code -- GPA calculation, for example, is duplicated, as well as the encoding of the rating-to-cost relationship (not sure if this second one can be helped or not).

natesholland:
Ok just dried out the code. I think at this point it's ready for a review from @guilhermesimoes, if you like it let me know and I can rebase it into a few (or one) pretty commit. If there are things you would like changed before they merge in let me know and I'll try to fix those things.

natesholland:
Added some tests too so now I'm not that guy.

guilhermesimoes:
Sorry, I was on a little vacation this weekend. I'll definitely take a look at your work this week.

lib/rubycritic/core/analysed_module.rb
```diff
@@ -27,6 +28,17 @@ def rating
...
+    def lines
+      return @lines if @lines
+      output = `wc -l < "#{path}"`.to_i
+      result = $?.success?
+      if result
+        @lines ||= output
+      else
+        @lines ||= File.open(path).lines.count
+      end
+    end
```
guilhermesimoes: `lines` was deprecated in favor of `each|each_line`.

I did a little [benchmark on Counting number of lines of a file in Ruby](https://gist.github.com/guilhermesimoes/d69e547884e556c3dc95) and `File.read(path).each_line.count` won.

However, `wc` was so slow that I'm thinking there's something wrong with it. Could you please run the benchmarks as well? If `wc` is not the fastest, I think we could simplify this to just using Ruby.

natesholland: Most certainly. I assumed that since wc was an optimized system function it would be faster, but I know spawning new system operations takes some overhead. If it turns out that wc is slower I agree that we should move to Ruby only.

lib/rubycritic/generators/global_rating_calculator.rb
```diff
@@ -0,0 +1,23 @@
+module Rubycritic
+  module Generator
+    module GlobalRatingCalculator
+      def calculate_gpa(analysed_modules)
+        gpa_sum = 0
+        total_lines = 0
+        analysed_modules.map do |analysed_module|
```
guilhermesimoes: Use `each` here.

guilhermesimoes:
@LeeXGreen Nice, forgot about `fetch`. But yeah, that `Configuration` class could be DRYed.

@natesholland Could you please move the threshold stuff to a separate PR? And please rebase over master and squash everything else.

So I've been thinking about weighting the GPA by line count and here's the thing. First, there are empty lines. And second, there's stuff like this:

```ruby
items.each do |item| 
  puts item
end
```
```ruby
items.each { |item| puts item }
```
Which should be weighted equally.

There's a reason people no longer use size metrics (like line counts) to measure software but instead use complexity metrics. I think in this case we could use our already computed complexity!

lib/rubycritic/core/analysed_module.rb
natesholland: I confirmed that the ruby version fast faster (about half the time on my tests: medium sized ruby project on MBP running Yosemite) so I moved it to use ruby only.

natesholland:
@guilhermesimoes when you say threshold stuff, do you mean you would like me to make one branch for the GPA threshold and one branch for the calculating and rendering of the GPA?

Also as for weighting by complexity, this sounds like a reasonable approach. My only concern with this approach is that I believe it would lead to bad classes affecting the GPA exponentially. I looked up code climate and they do their GPA calculations weighted on lines of code: http://docs.codeclimate.com/article/150-glossary-gpa. Maybe we could not do lines in the file but do something like lines in file minus empty lines minus comment lines?

guilhermesimoes:
> Do you mean you would like me to make one branch for the GPA threshold and one branch for the calculating and rendering of the GPA?

Yes, please. Let's keep this one to just calculating the GPA.

> My only concern with this approach is that I believe it would lead to bad classes affecting the GPA exponentially

Dang it, you might be right.

> Maybe we could not do lines in the file but do something like lines in file minus empty lines minus comment lines?

That's one possibility but I'm not sure of how easy that is. Maybe we can leave this like you originally coded it and revisit it later on. I would be interesting though if you tested a few variants of the "formula" on a Spiceworks project. At Whitesmith we could do the same.

natesholland:
Ok see #34 for merging in the gpa calculation and rendering only. Then I'll go back and rethink how to do the threshold stuff.

natesholland:
I changed my master branch to be upstream master and moved all these features onto their own branch called `gpa-and-threshold`

PR #33, opened 2015-03-09
Design improvements

crackofdusk:
rubycritic-design-improvements

Theses changes improve the readability of sortable tables. They:
    
  * add arrows indicating an ascending or descending sort order
  * let the browser compute column widths (columns with longer content are larger)
  * highlight the row under the cursor
  * set a maximum page width
  * distinguish the header from the rest of the content

I also want to add better default sorting where it makes sense (new code smells on top, files with worse rating on top), but that requires some cleaning up the JavaScript code first. I will do that in another pull request in the future if this one gets accepted.


guilhermesimoes:
Oh wow, this looks so much better! I'll definitely take a look at your work this week.

guilhermesimoes:
I played a little with your branch and the project looks so much better!

I couldn't help but notice that the table header would look even better with some bottom shadow to hide the top row that is clipped. Something like:

```
.floatThead-container {
  box-shadow: 0px 4px 4px -2px lightgrey;
}
```

You most certainly can improve that as well! :laughing:

crackofdusk:
Thanks for going through the code. I'm usually not that sloppy — lack of sleep got the best of me there.

I cleaned it up and also added a subtle shadow like you suggested.

Let me know when you want me to squash the commits.

guilhermesimoes:
Looks good to me! Please squash everything except the tablesorter version update.

And this is just me venting but I would really like it if we didn't have to include these huge JavaScript files in the GitHub repo. I wish we could use Bower but I think that is out of the question. Having Node as an extra dependency would be yet another obstacle to newcomers who want to contribute...

crackofdusk:
I squashed the commits.

guilhermesimoes:
Merged!

Meanwhile, some recent changes I made to `rubycritic` have made it incompatible with `guard-rubycritic`. I will fix that this weekend and then release a new version.

Thanks for contributing! :smiley:

PR #34, opened 2015-03-14
adding the calculation and rendering of gpa weighted by number of lines per file

natesholland:
Ok I rebased this all into one commit so it should be ready to come in. If you see anything you don't like please let me know.

### DIFF
--- a/lib/rubycritic/core/analysed_module.rb
+++ b/lib/rubycritic/core/analysed_module.rb
@@ -29,7 +29,12 @@ module Rubycritic
     end
 
     def lines
-      @lines ||= File.open(path).each_line.count
+      return @lines if @lines
+      count = 0
+      File.open(path).each_line do |line|
+        count += 1 unless line.strip[0] == "#" || line.strip.empty?
+      end
+      @lines = count
     end
 
     def complexity_per_method
--- a/lib/rubycritic/core/analysed_module.rb
+++ b/lib/rubycritic/core/analysed_module.rb
@@ -14,6 +14,7 @@ module Rubycritic
     attribute :complexity
     attribute :duplication,   Integer, :default => 0
     attribute :methods_count
+    attribute :lines
 
     def path
       @path ||= pathname.to_s
@@ -25,6 +26,10 @@ module Rubycritic
 
     def rating
       @rating ||= Rating.from_cost(cost)
+    end
+
+    def lines
+      @lines ||= File.open(path).each_line.count
     end
 
     def complexity_per_method
--- a/lib/rubycritic/core/rating.rb
+++ b/lib/rubycritic/core/rating.rb
@@ -25,6 +25,17 @@ module Rubycritic
     def to_json(*a)
       to_h.to_json(*a)
     end
+
+    def to_gpa
+      if @letter == "A" then 4.0
+      elsif @letter == "B" then 3.0
+      elsif @letter == "C" then 2.0
+      elsif @letter == "D" then 1.0
+      elsif @letter == "F" then 0.0
+      else
+        raise "unknown GPA"
+      end
+    end
   end
 
 end
--- a/lib/rubycritic/generators/global_rating_calculator.rb
+++ b/lib/rubycritic/generators/global_rating_calculator.rb
@@ -1,7 +1,7 @@
 module Rubycritic
   module Generator
-    module GlobalRatingCalculator
-      def calculate_gpa(analysed_modules)
+    class GlobalRatingCalculator
+      def self.calculate_gpa(analysed_modules)
         gpa_sum = 0
         total_lines = 0
         analysed_modules.each do |analysed_module|
--- a/lib/rubycritic/generators/global_rating_calculator.rb
+++ b/lib/rubycritic/generators/global_rating_calculator.rb
@@ -1,0 +1,15 @@
+module Rubycritic
+  module Generator
+    module GlobalRatingCalculator
+      def calculate_gpa(analysed_modules)
+        gpa_sum = 0
+        total_lines = 0
+        analysed_modules.each do |analysed_module|
+          gpa_sum += (analysed_module.rating.to_gpa) * analysed_module.lines
+          total_lines += analysed_module.lines
+        end
+        gpa_sum / total_lines
+      end
+    end
+  end
+end
--- a/lib/rubycritic/generators/html_report.rb
+++ b/lib/rubycritic/generators/html_report.rb
@@ -9,12 +9,12 @@ module Rubycritic
   module Generator
 
     class HtmlReport
-      include GlobalRatingCalculator
+
       ASSETS_DIR = File.expand_path("../html/assets", __FILE__)
 
       def initialize(analysed_modules)
         @analysed_modules = analysed_modules
-        @gpa = calculate_gpa(analysed_modules)
+        @gpa = GlobalRatingCalculator.calculate_gpa(analysed_modules)
       end
 
       def generate_report
--- a/lib/rubycritic/generators/html_report.rb
+++ b/lib/rubycritic/generators/html_report.rb
@@ -3,15 +3,18 @@ require "rubycritic/generators/html/overview"
 require "rubycritic/generators/html/smells_index"
 require "rubycritic/generators/html/code_index"
 require "rubycritic/generators/html/code_file"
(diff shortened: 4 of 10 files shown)


lib/rubycritic/core/analysed_module.rb
```diff
@@ -14,6 +14,7 @@ class AnalysedModule
     attribute :complexity
     attribute :duplication,   Integer, :default => 0
     attribute :methods_count
+    attribute :lines
```
guilhermesimoes: Must have missed this before, but I believe this line is not needed.

lib/rubycritic/generators/global_rating_calculator.rb
```diff
@@ -0,0 +1,15 @@
+module Rubycritic
+  module Generator
+    module GlobalRatingCalculator
+      def calculate_gpa(analysed_modules)
+        gpa_sum = 0
+        total_lines = 0
+        analysed_modules.each do |analysed_module|
+          gpa_sum += (analysed_module.rating.to_gpa) * analysed_module.lines
```
guilhermesimoes: What's the reasoning for converting the rating back to a GPA step? I think I like #35's approach better, averaging the costs of the modules and then using the Rating class to convert it to a GPA.

lib/rubycritic/generators/html_report.rb
```diff
@@ -3,15 +3,18 @@
...
 module Rubycritic
   module Generator
 
     class HtmlReport
+      include GlobalRatingCalculator
       ASSETS_DIR = File.expand_path("../html/assets", __FILE__)
 
       def initialize(analysed_modules)
         @analysed_modules = analysed_modules
+        @gpa = calculate_gpa(analysed_modules)
```
guilhermesimoes: I don't think it's a `Report`'s responsibility to know how to calculate the GPA. Make `GlobalRatingCalculator` a class and call this method on it!

lib/rubycritic/generators/global_rating_calculator.rb
natesholland: I think converting the rating back to a GPA makes sense because it will put a cap on how much a really bad file can affect the total GPA. Since the rating can have an arbitrarily high cost it means that when averaged together a really bad F can skew the entire rating. I know at Spiceworks we have a model that is very very bad (over 4000 lines of code) so it would skew our GPA lower than if we made all F's the same.

lib/rubycritic/core/analysed_module.rb
natesholland: This line is needed if we want to have the lines= method which is very useful in testing. I can pull it out if you want and then rework the tests, but I think it's worth leaving in for testing purposes.

lib/rubycritic/generators/html_report.rb
natesholland: Agreed! And done.

natesholland:
It turned out that making the weighting by lines of code seemed to have little effect on the total time needed to calculate the GPA, so I went ahead and weighted by lines of code instead of lines in the file.

natesholland:
@guilhermesimoes can you pick this or #35 and merge one of them in? I think there are enough people who want these changes to warrant merging them in.

natesholland:
@guilhermesimoes any more thoughts on this? I really don't want to fork this internally at Spiceworks but I am probably going to if we can't pick some form of internal global rating. I'd be happy to do extra work on this to get it merged in if there is anything you are waiting on. Please let me know if you intent to merge this in.

JustinAiken:
:+1: for picking one - no opinion as to which one.

natesholland:
@Onumis and @troessner can we get this merged into the main trunk of rubycritic?

troessner:
@natesholland i have "review all open pull requests" on my to do list already. Please give me / us some time to become fully acquainted with the project.

natesholland:
Sounds good, I just wanted to make sure it was in the queue

lib/rubycritic/core/analysed_module.rb
```diff
@@ -27,6 +28,15 @@ def rating
       @rating ||= Rating.from_cost(cost)
     end
 
+    def lines
```
Onumis: `lines_of_code` would be a more appropriate name for this.

Onumis:
Ok so, I'm gonna have to lay some though on this and #35, there are a lot of really interesting opinions here and there and I'm not convinced on which one is the "right path".

lib/rubycritic/core/analysed_module.rb
```diff
@@ -27,6 +28,15 @@ def rating
...
+      return @lines if @lines
```
troessner: You could use the more idiomatic memoization pattern:

```Ruby
@lines ||= begin
  .....
end
```

lib/rubycritic/core/analysed_module.rb
```diff
@@ -27,6 +28,15 @@ def rating
...
+      count = 0
+      File.open(path).each_line do |line|
+        count += 1 unless line.strip[0] == "#" || line.strip.empty?
```
troessner: Let's move the guard

```Ruby
line.strip[0] == "#" || line.strip.empty?
```

into an appropriately named method.
How about

```Ruby
def ignore_line?(line)
  line.strip[0] == "#" || line.strip.empty?
end
```

?

You could directly use File's `readlines` and Enumerable's `reject`, then this whole method could be written more concise like this:

```Ruby
def lines
  @lines ||= File.open(path).readlines
                                 .reject { |line| ignore_line?(line) }
                                  .size
```

Then you can also drop the "begin" I suggested above ;)

lib/rubycritic/core/rating.rb
```diff
@@ -25,6 +25,17 @@ def to_h
     def to_json(*a)
       to_h.to_json(*a)
     end
+
+    def to_gpa
+      if @letter == "A" then 4.0
+      elsif @letter == "B" then 3.0
+      elsif @letter == "C" then 2.0
+      elsif @letter == "D" then 1.0
+      elsif @letter == "F" then 0.0
```
troessner: Let's not mix up code with data.
How about putting the association of character to number into a hash?

```Ruby
{ 'A' => 4.0, ....}
````
Then you could just use Hash#fetch on that.

File: lib/rubycritic/generators/global_rating_calculator.rb [hunk 8]
```diff
@@ -0,0 +1,15 @@
+module Rubycritic
+  module Generator
+    class GlobalRatingCalculator
```
troessner: You never plan to instantiate this, right?
Then let's rather use a module and `module_function` instead of self :)

test/lib/rubycritic/report_generators/global_rating_calculator_test.rb
```diff
@@ -0,0 +1,15 @@
...
+require "rubycritic/core/analysed_module"
+
+describe Rubycritic::Generator::GlobalRatingCalculator do
+  it "calculates the gpa" do
+    smells1 = [SmellDouble.new(:cost => 1), SmellDouble.new(:cost => 2)]
+    smells2 = [SmellDouble.new(:cost => 10), SmellDouble.new(:cost => 12)]
+    am1 = Rubycritic::AnalysedModule.new(:smells => smells1, :complexity => 0, :lines => 25)
+    am2 = Rubycritic::AnalysedModule.new(:smells => smells2, :complexity => 0, :lines => 10)
+    analysed_modules = [am1, am2]
+    gpa = Rubycritic::Generator::GlobalRatingCalculator.calculate_gpa(analysed_modules)
```
troessner: The test set up here is huge. Normally I'd suggest to introduce factories here but I guess this would be a bigger topic that should be adressed in an separate pull request.

lib/rubycritic/generators/global_rating_calculator.rb
```diff
@@ -0,0 +1,15 @@
...
+      def self.calculate_gpa(analysed_modules)
+        gpa_sum = 0
+        total_lines = 0
+        analysed_modules.each do |analysed_module|
+          gpa_sum += (analysed_module.rating.to_gpa) * analysed_module.lines
```
troessner: The braces dont make sense here, seems like a leftover?

troessner:
Closing this one since we merged #35 - thanks!

PR #35, opened 2015-03-15
Experiments for adding a global score/rating

ancorgs:
In https://github.com/whitesmith/rubycritic/issues/9#issuecomment-77242630 it was suggested to open a pull request to start discussing possible alternatives to implement global rating/scoring for a whole project/repository (ala CodeClimate).

It makes sense to me to just use the average cost of all analyzed modules and present that number in some meaningful way, either as a rating or as a number between 5 (perfect) and 0 (crap).

Since in this PR I want to focus the discussion on the way of turning the average cost into something meaningful I did almost no intrusive changes to the existing code apart from adding some ```puts``` so we can see the values without changing the reports yet.

After preparing this PR, I saw #32. I have not gone through its code yet, but it seems to include a very well though mechanism to achieve the same goal, so maybe we are already close enough to have global ratings in rubycritic and this PR is useless. If it's the case, don't hesitate to close it right away.

### DIFF
--- a/lib/rubycritic/analysers_runner.rb
+++ b/lib/rubycritic/analysers_runner.rb
@@ -1,5 +1,4 @@
-require "rubycritic/source_locator"
-require "rubycritic/core/analysed_module"
+require "rubycritic/core/analysed_modules_collection"
 require "rubycritic/analysers/smells/flay"
 require "rubycritic/analysers/smells/flog"
 require "rubycritic/analysers/smells/reek"
@@ -29,9 +28,7 @@ module Rubycritic
     end
 
     def analysed_modules
-      @analysed_modules ||= SourceLocator.new(@paths).pathnames.map do |pathname|
-        AnalysedModule.new(:pathname => pathname)
-      end
+      @analysed_modules ||= AnalysedModulesCollection.new(@paths)
     end
   end
 
--- a/lib/rubycritic/core/analysed_modules_collection.rb
+++ b/lib/rubycritic/core/analysed_modules_collection.rb
@@ -1,0 +1,49 @@
+require "rubycritic/source_locator"
+require "rubycritic/core/analysed_module"
+
+module Rubycritic
+  class AnalysedModulesCollection
+    include Enumerable
+
+    # Limit used to prevent very bad modules to have excessive impact in the
+    # overall result. See #limited_cost_for
+    COST_LIMIT = 32
+    # Score goes from 0 (worst) to 100 (perfect)
+    MAX_SCORE = 100
+    # Projects with an average cost of 16 (or above) will score 0, since 16
+    # is where the worst possible rating (F) starts
+    ZERO_SCORE_COST = 16
+    COST_MULTIPLIER = MAX_SCORE.to_f / ZERO_SCORE_COST
+
+    def initialize(paths)
+      @modules = SourceLocator.new(paths).pathnames.map do |pathname|
+        AnalysedModule.new(:pathname => pathname)
+      end
+    end
+
+    def each(&block)
+      @modules.each(&block)
+    end
+
+    def to_json(*options)
+      @modules.to_json(*options)
+    end
+
+    def score
+      MAX_SCORE - average_limited_cost * COST_MULTIPLIER
+    rescue
+      0.0
+    end
+
+    private
+
+    def average_limited_cost
+      avg = map { |mod| limited_cost_for(mod) }.reduce(:+) / @modules.size.to_f
+      [avg, ZERO_SCORE_COST].min
+    end
+
+    def limited_cost_for(mod)
+      [mod.cost, COST_LIMIT].min
+    end
+  end
+end
--- a/lib/rubycritic/generators/html/overview.rb
+++ b/lib/rubycritic/generators/html/overview.rb
@@ -10,6 +10,8 @@ module Rubycritic
 
         def initialize(analysed_modules)
           @turbulence_data = Turbulence.data(analysed_modules)
+          @score = analysed_modules.score
+          @max_score = AnalysedModulesCollection::MAX_SCORE
         end
 
         def file_name
--- a/lib/rubycritic/generators/html/templates/overview.html.erb
+++ b/lib/rubycritic/generators/html/templates/overview.html.erb
@@ -1,3 +1,7 @@
+<div class="chart-container">
+  <h2>Score: <%= @score.round(2) %> / <%= @max_score %></h2>
+</div>
+
 <div id="js-chart-container" class="chart-container"></div>
 
 <script type="text/javascript">
--- a/lib/rubycritic/generators/json/simple.rb
+++ b/lib/rubycritic/generators/json/simple.rb
@@ -20,7 +20,8 @@ module Rubycritic
                 :version => Rubycritic::VERSION
               }
             },
-            :analysed_modules => @analysed_modules
+            :analysed_modules => @analysed_modules,
+            :score => @analysed_modules.score
           }
         end
       end
--- a/test/lib/rubycritic/core/analysed_modules_collection_test.rb
+++ b/test/lib/rubycritic/core/analysed_modules_collection_test.rb
@@ -1,0 +1,109 @@
+require "test_helper"
+require "rubycritic/core/analysed_modules_collection"
+
+describe Rubycritic::AnalysedModulesCollection do
+  subject { Rubycritic::AnalysedModulesCollection.new(paths) }
+
+  describe ".new" do
+    context "with an empty path" do
+      let(:paths) { "" }
+
+      it "returns an empty collection" do
+        subject.count.must_equal 0
+      end
+    end
+
+    context "with a list of files" do
+      let(:paths) { %w(test/samples/doesnt_exist.rb test/samples/unparsable.rb test/samples/empty.rb) }
+
+      it "registers one AnalysedModule element per existent file" do
+        subject.count.must_equal 2
+        subject.all? { |a| a.is_a?(Rubycritic::AnalysedModule) }.must_equal true
+      end
+    end
+
+    context "with a directory" do
+      let(:paths) { "test/samples/" }
+
+      it "recursively registers all files" do
+        subject.count.must_equal 13
+      end
+    end
+
+    context "with redundant paths" do
+      let(:paths) { %w(test/samples/flog test/samples/flog/complex.rb) }
+
+      it "returns a redundant collection" do
+        subject.count.must_equal 3
+      end
+    end
+  end
+
+  describe "#score" do
+    context "with no modules" do
+      let(:paths) { "" }
+
+      it "returns zero" do
+        subject.score.must_equal 0.0
+      end
+    end
+
+    context "with not analysed modules" do
+      let(:paths) { "test/samples/flog" }
+
+      it "returns zero" do
+        subject.score.must_equal 0.0
+      end
+    end
+
+    context "with analysed modules" do
+      before do
+        subject.each_with_index do |mod, i|
+          mod.expects(:cost).returns costs[i]
+        end
+      end
+
+      let(:paths) { %w(test/samples/flog test/samples/flay) }
+
+      context "with perfect modules" do
+        let(:costs) { [0.0, 0.0, 0.0, 0.0] }
+
+        it "returns the maximum score" do
+          subject.score.must_equal 100.0
+        end
+      end
+
+      context "with very bad modules" do
+        let(:costs) { [16.0, 16.0, 16.0, 16.0] }
+
+        it "returns zero" do
+          subject.score.must_equal 0.0
+        end
+      end
+
+      context "with horrible modules" do
+        let(:costs) { [32.0, 32.0, 32.0, 32.0] }
+
+        it "returns zero" do
+          subject.score.must_equal 0.0
+        end
+      end
+
+      context "with mixed modules" do
+        let(:costs) { [32.0, 2.0, 0.0, 2.0] }
+
+        it "properly calculates the score" do
+          subject.score.must_equal 43.75
+        end
+      end
+
+      context "with a module above the cost limit" do
+        let(:costs) { [220.0, 2.0, 0.0, 2.0] }
+
+        it "reduces the impact in the result" do
+          subject.score.must_equal 43.75
+        end
+      end
+    end
+  end
+end


natesholland:
Hopefully #34 get's pulled in soon. One thing I would like your input on is how do we weight GPA by? Each file gets the same weight, lines in the file, lines of code in the file, complexity of the file? Any thoughts on these?

guilhermesimoes:
Ooh, nice. I really like this approach, super clean! Tomorrow we're going to play with these global ratings to get a better feel of what they represent.

natesholland:
I really like the implementation here (very clean) but I disagree with using the average cost. My reasoning behind this is that using the average cost will allow a few files with a very high cost to skew the average towards a bad over all rating. For example we have a model with a cost of 220. Since the threshold for an F is 16 this means that we have models that will drastically pull down our average and probably give the project an average GPA of an F. Soon I will look in my branch and try these ways of calculating the average GPA on three of our projects at Spiceworks and report back with my findings.

ancorgs:
@natesholland I see your point and surely you are right. But the approach used in your proposal of converting a number (the cost) to letters and then back to numbers (GPA) looks like a lot of rounding to me. A lot of precision is lost. The same effect (without the precision lost) can be achieved filtering the extreme values while calculating the average, something like:
```ruby
class AnalysedModule
  def limited_cost
    cost < 25 ? cost : 25
  end
end

class AnalysedModulesCollection
  def average_limited_cost
    self.map(&:limited_cost).reduce(:+) / @modules.size.to_f
  end

  def sqrt_score
    5 - Math.sqrt(average_limited_cost)
  end
end
```

ancorgs:
BTW, looking closely I think that the main weak point in this pull request is assuming that ```Rating.from_cost``` is actually implementing some kind of rounded square root. Looking at the code, my guts feelings say me that it should be easily translatable to a mathematical function, even if it's certainly not _round(sqrt(x))_ as I wrongly assumed when suggesting ```sqrt_score``` as a good measure.

ancorgs:
I realized that the method that used to be called ```sqrt_score``` was not the correct inverse of ```Rating.from_cost```. The inverse of ```2**x``` is actually ```log2(x)``` and not ```sqrt(x)```. Thus, I updated the pull request. I also included a change to reduce the overall impact of very nasty modules, as pointed by @natesholland.

natesholland:
Ok I've dived into testing this on some different projects so I'll present my data and then my conclusions and then some suggestions for where we should go from here.

First I decided to break this down into the three rating calculation methods that I found most compelling. From here on out let us call the traditional GPA calculations I did in #34 option 1. Then let's call the non-linear average cost option 2, and the linear average cost option 3. (these match up with the comments in the code.)

## Data

I ran these on two of our projects here at spiceworks. I would describe one (project A) as a small/medium project and the other (project B) as a medium sized project. Both have decent code quality but certainly have their C's and D's (though no F's in either one). For reference, Project A has 2 D's 6 C's and 2 B's with the rest being A's, where Project B has 10 D's 16C's and 17 B's with the rest being A's.
So here is the data from running option 1 on both projects.

### Option 1

screen shot 2015-03-16 at 8 58 37 am
screen shot 2015-03-16 at 8 58 44 am

As you can see from the scatter plot project B is a good bit larger than project A and there are more files with some issues. The GPA seems to reflect this and show that both projects are doing pretty good but not awesome. Project A's GPA is roughly a B, where Project B's GPA is roughly a B- (according to how most American schools would rate this. Granted, as an American, I may be biased and full of myself and all that).

### Option 2

screen shot 2015-03-16 at 9 04 11 am
screen shot 2015-03-16 at 9 04 30 am

Interestingly in option 2 the overall rating for the two projects came out to be almost identical. I own project A so I showed these comparisons to the owner of Project B and we both agreed that the human assessment of our projects is that they should have different over all ratings on code quality. (side note, his project makes more money for the company and code quality isn't everything, etc.).

### Option 3

screen shot 2015-03-16 at 9 05 49 am
screen shot 2015-03-16 at 9 06 04 am

In option 3 the results were similar to option 2. Both projects' scores were close together but the linear option seems to have inflated the score even further.

 ## Conclusions

I went ahead and showed this to a senior developer here at Spiceworks and he and I came to the same conclusion that we like option 1 the most and here was our reasoning. We thought that Project A and Project B should have different over all ratings and project B should be lower than project A, though not considerably. What we think is happening in Option 2 and 3 is that since they are not weighting the files by size a bunch of small A files are masking the few large D files. Furthermore in a Rails project it's really common to have a few big nasty files (the user model, application model, and/or domain model) and we think that those files should have a weight on the over all rating of the project. I researched how code climate is doing this and they have chosen to go with option 1 per their documentation [here.](http://docs.codeclimate.com/article/150-glossary-gpa)

## Suggestions

Where do we go from here? I certainly understand the concern of the rounding error and the loss of granularity that that causes.  It's a common frustration among students in the school system. One option I think we have, though I'm not convince we should take, is to introduce the concept of plusses and minuses in the grades. For example, at UC Berkeley we had a GPA scale where A => 4.0, A- => 3.7, B+ => 3.3, b => 3.0 and so on. At this point I would like to hear feedback from @guilhermesimoes @ancorgs @LeeXGreen and @robmack and anyone else who's interested in the project.

natesholland:
Ok one more data point. I ran all the options on one of our biggest and oldest projects at spiceworks and here were the results:

* Option 1 (GPA): 1.34
* Option 2 (log2): 3.12
* Option 3 (linear): 4.27

This project has over 100 F's and I think it should definitely have a low over all rate.

ancorgs:
@natesholland I decided to do a similar experiment running the code on this PR against several open source projects available at Code Climate. Here are the results.

**Non-Rails applications**

https://github.com/yast/yast-journal (CodeClimate 3.5)
The (limited) average cost is: 3.3529411764705883
which means a rating of: B
which means a nonlinear_score of: 3.77 over 5
which means a linear_score of: 4.48 over 5

https://github.com/yast/yast-network (CodeClimate 1.0)
The (limited) average cost is: 9.693181818181818
which means a rating of: D
which means a nonlinear_score of: 1.88 over 5
which means a linear_score of: 3.49 over 5

https://github.com/yast/yast-registration (CodeClimate 2.4)
The (limited) average cost is: 5.913793103448276
which means a rating of: C
which means a nonlinear_score of: 2.70 over 5
which means a linear_score of: 4.08 over 5

**Partially Rails application**

https://github.com/openSUSE/open-build-service (CodeClimate 3.4)
The (limited) average cost is: 4.626666666666667
which means a rating of: C
which means a nonlinear_score of: 3.14 over 5
which means a linear_score of: 4.28 over 5

**Pure Rails applications**

https://github.com/openSUSE/osem (CodeClimate 2.2)
The (limited) average cost is: 2.4270557029177717
which means a rating of: A
which means a nonlinear_score of: 4.19 over 5
which means a linear_score of: 4.62 over 5

https://github.com/openSUSE/travel-support-program (CodeClimate 1.6)
The (limited) average cost is: 3.2228915662650603
which means a rating of: B
which means a nonlinear_score of: 3.85 over 5
which means a linear_score of: 4.50 over 5

As you can see, the results for the non-linear score look more promising than in your tests, but this PR is way too generous with the Rails applications for some reason. In the last case, travel-support-program, the Javascript code has a huge influence, since many of the files rated low by CodeClimate are actually JS (and ignored by rubycritic). Even though, the difference between CodeClimate and rubycritic is still too high even for OSEM, which does not suffer from the Javascript factor.

I will now do the same test with the other approach to compare results and try to investigate the problem with Rails (maybe the weighting introduced in #34 helps).

ancorgs:
Same repositories using PR#34 (GPA).

**Non-Rails applications**

https://github.com/yast/yast-journal (CodeClimate 3.5)
GPA: 2.60

https://github.com/yast/yast-network (CodeClimate 1.0)
GPA: 0.97

https://github.com/yast/yast-registration (CodeClimate 2.4)
GPA: 1.86

**Partially Rails application**

https://github.com/openSUSE/open-build-service (CodeClimate 3.4)
GPA: 0.95

**Pure Rails applications**

https://github.com/openSUSE/osem (CodeClimate 2.2)
GPA: 2.00

https://github.com/openSUSE/travel-support-program (CodeClimate 1.6)
GPA: 2.35

It always rates much lower than the alternative in this PR and it always rate slightly lower than CodeClimate. The two cases in which the GPA differs a lot from CodeClimate can be easily explained (just a theory though) by the presence of Javascript. The Ruby code in travel-support-program is quite ok, but the JS sucks. In open-build-service is exactly the other way around. So the difference fits my expectations.

So in general the GPA looks okish, but I find too hard to reach a 3 (only 2 projects above 2 with a maximum of 2.60). That can be frustrating.

ancorgs:
You know what? I was wrongly assuming that the CodeClimate's and #34's GPA go from 0 to 5 instead of 1 to 4 (something that can be obvious and needles to say for US citizen, but different in every part of the world). So the whole comparison I have done looks quite silly actually :-P

So first question is. Do we want to rate everything using the 4.0 scale explained here?
http://www.collegeboard.com/html/academicTracker-howtoconvert.html

If that's the case, I'd need to adjust my approach. Otherwise, comparing to CodeClimate or #34 is pointless.

ancorgs:
In fact, if we want to go the GPA way, adjusting this PR makes no sense and we should simply go with #34 (although I'm not completely sure if weighting by line count of the file really makes sense, specially counting lines instead of real lines of code).

Adjusting this PR to the GPA 4.0 scale makes no sense because both systems are fundamentally different. I was looking for the maths behind the ratings in search for precision. The GPA 4.0 scale system is more arbitrary (no offense meant, it's perfectly fine), based in tables and conventions instead of mathematical functions.

Not having being exposed to a rating system based in letters and a table-based GPA, my approach looks more natural to me (we simply use a linear system from 0 to 10 in my country). But I'm obviously biased.

natesholland:
@ancorgs thanks for running all those calculations. I totally agree that we should weight by lines of code, not lines in the file. I am going to look into how to do that quickly, just haven't had time for that yet.

As for whether we should do the American GPA or something closer to what you described I have no idea. I tried looking through this wikipedia article http://en.wikipedia.org/wiki/Grading_systems_by_country but as I perused through it I couldn't find any consistent standard amongst the countries. North America, China, Hong Kong, and parts of South America follow the table based GPA scale that I'm used to. India and the Netherlands do a 1-10 scale where bigger is better, but South Korea does 1-10 where smaller is better. Even amongst neighboring European countries (where I would have expected standardization) Germany does a 1-5 where smaller is better, France does 1 to 20, Spain does 1-10, and Portugal does 1-5 all where bigger is better.

All that is to say I have no idea which one is the most widely adopted or easily understood. This is in no way a standard v. metric type of discussion (where metric is the most commonly adopted and most easily understood). Since I was raised in the American system I don't think I can speak about this without a bias. I was hoping there would be a general consensus amongst the international community, which would make this easier, but that does not seem to be the case.

ancorgs:
@guilhermesimoes any plan to move this forward? I think we have enough information, insights and example data to select one approach (and the one in #34 seems to be the more codeclimate-alike) and move forward.

doomspork:
This feature would be really awesome to have.  Using @natesholland's `gpa_calc_only` branch and [shields.io](http://shields.io) I set up a Jenkins instance to comment on each PR with a code quality badge much like Code Climate does:

Quality
Quality
Quality
Quality
Quality
Quality

I'd love to get involved if that would help move this forward.

Thanks again for an awesome tool :+1:

troessner:
I'll take over this branch and will review it in the next couple of days ;)

squaresurf:
@troessner do you have any updates on this? I really want to get a badge like @doomspork is showing on our repos.

troessner:
Nope, no updates right now, but this PR is next on my RubyCritic todo list, promised.

squaresurf:
Awesome! Thank you.

lib/rubycritic/core/analysed_module.rb
```diff
@@ -23,6 +23,10 @@ def cost
       @cost ||= smells.map(&:cost).inject(0, :+) + (complexity / 25)
     end
 
+    def limited_cost
+      cost < 32 ? cost : 32
```
troessner: Magic number - why 32? Literal duplication as well. How about making this a constant?

lib/rubycritic/core/analysed_modules_collection.rb
```diff
@@ -0,0 +1,54 @@
...
+        5 - cost/3
+      end
+    end
+
+    # Option 3: from 5 to 0 (CodeClimate's style) by just applying a linear
+    # function to the cost with a cost of 32 being a 0 and a cost of 0 meaning,
+    # of course, a score of 5.
+    # This way, the progression is smoother than the one based on ratings.
+    def linear_score
+      score = 5 - average_cost / 6.4
```
troessner: 6.4? Magic number again.

lib/rubycritic/analysers_runner.rb
```diff
@@ -25,13 +24,15 @@ def initialize(paths)
 
     def run
       ANALYSERS.each { |analyser| analyser.new(analysed_modules).run }
+      puts "The (limited) average cost is: #{analysed_modules.average_cost}"
+      puts "which means a rating of: #{analysed_modules.rating}"
+      puts "which means a nonlinear_score of: %.2f over 5" % analysed_modules.nonlinear_score
+      puts "which means a linear_score of: %.2f over 5" % analysed_modules.linear_score
```
troessner: Lets drop the "which means" everywhere, dont think we need it.

lib/rubycritic/core/analysed_modules_collection.rb
```diff
@@ -0,0 +1,54 @@
+require "rubycritic/source_locator"
+require "rubycritic/core/analysed_module"
+
+module Rubycritic
+
```
troessner: Can we drop the newline here?

lib/rubycritic/core/analysed_modules_collection.rb
```diff
@@ -0,0 +1,54 @@
...
+    def each(&block)
+      @modules.each(&block)
+    end
+
+    def to_json(*a)
+      @modules.to_json(*a)
+    end
+
+    def average_cost
+      self.map(&:limited_cost).reduce(:+) / @modules.size.to_f
```
troessner: The "self" is not needed,  lets drop it.

troessner:
I'd prefer this solution over #34
A couple of small improvements and I'd vote for merging this ;)

One more thing: can you please summarize the rationale for the calculations in the readme in this PR? Basically a short version of your comments above ;)

troessner:
@ancorgs and since I forgot to mention it, nice job!
@lucasmazza & @Onumis what do you think?

lib/rubycritic/analysers_runner.rb
```diff
@@ -25,13 +24,15 @@ def initialize(paths)
 
     def run
       ANALYSERS.each { |analyser| analyser.new(analysed_modules).run }
+      puts "The (limited) average cost is: #{analysed_modules.average_cost}"
```
lucasmazza: Similar feelings from https://github.com/whitesmith/rubycritic/pull/36#issuecomment-158670848 - the output shouldn't be inside the `AnalysersRunner` class.

lib/rubycritic/analysers_runner.rb
lucasmazza: :+1:

lucasmazza:
> One more thing: can you please summarize the rationale for the calculations in the readme in this PR? Basically a short version of your comments above ;)

:+1: it would also be helpful to document the calculations in the README or the wiki for future users.

lib/rubycritic/analysers_runner.rb
lucasmazza: Maybe we should merge these PR as is and refactor the existing reports with these pieces elsewhere so we don't stall the existing PRs too much :)

ancorgs:
As stated in the description, this PR was never intended to merge as is, but as a base for discussion. Thus the usage of `puts`, the magic numbers and all that.

I was away from the computer last week. I'm back now and will try to make this production ready. Thanks for the comments.

ancorgs:
This PR included three alternatives for the calculation, labeled as:

* Option 1 - rating
* Option 2 - nonlinear score - from 5 to 0 converting the rating back to numbers
* Option 3 - linear_score - from 5 to 0 but just applying a linear progression

Which one do you want to be used finally?

ancorgs:
Now that I really understand how the GPA (a table with letters from A to F) system works and that I know that CodeClimate rates from 1 to 4 (and not 0 to 5 as I originally thought), I will probably adjust this PR to use the linear approach, since it's more smooth and easier to understand.

Onumis:
Indeed. I think a `linear_score` is the best option. We really want to keep it simple and easy to understand.

troessner:
>> Indeed. I think a linear_score is the best option. We really want to keep it simple and easy to understand.

Agreed. Let's apply KISS here and not get fancy.

ancorgs:
I squashed everything into a single commit and updated this PR. Please review.

ancorgs:
Well, let me please rubocop before reviewing :wink:

ancorgs:
Offenses fixed. Ready for review.

natesholland:
This looks good to me :+1: 

Is there anyway we can cut a 1.9 branch that I can maintain? Here at Spiceworks we run a lot of 1.9 ruby and I would like to cherry pick this back onto a 1.9 branch so we can use this feature.

troessner:
>> Offenses fixed. Ready for review.

But now this is a) nowhere used and b) doesn't have any tests?

>> Is there anyway we can cut a 1.9 branch that I can maintain? 

That would be awesome! I'd happily create this branch for you. From which commit would you like to start?
This one:

>> 10d8c5c - Update RuboCop so that its Parser version is compatible with Reek 3 (13 days ago) <Guilherme Simoes>

ancorgs:
I will add some tests now. And will also try to make it fit in the templates, but don't expect anything fancy :-)

ancorgs:
Added tests, added score to the templates and fixed some bugs found while writing the tests.

Everything squashed into a single commit.

Now I dare to say it's ready to be merged.

troessner:
Awesome job! Merged - thanks ;)

Onumis:
:heart:

troessner:
Let's get #83 merged and then release a new version so we can get this feature out ;)

PR #36, opened 2015-03-16
first thoughts on progress indication in rubycritic

natesholland:
I'm starting with some rough ideas on how to do progress indication in rubycritic. I know the code should be dried out and all that but I mainly want to begin discussing how we should present the information to the user. My first stab at it I showed an indication bar for each analyzer, (flay, flog, churn...) because that's how the code is currently organized. It seems more intuitive to me to show it file by file, but that's not currently how we are iterating through things so it's a bit counter intuitive in the code.

Here's what my first stab at it looks like:
screen shot 2015-03-16 at 11 50 36 am

### DIFF
--- a/lib/rubycritic/analysers/smells/flog.rb
+++ b/lib/rubycritic/analysers/smells/flog.rb
@@ -1,10 +1,12 @@
 require "rubycritic/analysers/helpers/flog"
 require "rubycritic/core/smell"
+require "rubycritic/colorize"
 
 module Rubycritic
   module Analyser
 
     class FlogSmells
+      include Colorize
       HIGH_COMPLEXITY_SCORE_THRESHOLD = 25
       VERY_HIGH_COMPLEXITY_SCORE_THRESHOLD = 60
 
@@ -16,7 +18,13 @@ module Rubycritic
       def run
         @analysed_modules.each do |analysed_module|
           add_smells_to(analysed_module)
+          print green "."
         end
+        puts ""
+      end
+
+      def to_s
+        "flog smells"
       end
 
       private
--- a/lib/rubycritic/analysers_runner.rb
+++ b/lib/rubycritic/analysers_runner.rb
@@ -24,7 +24,11 @@ module Rubycritic
     end
 
     def run
-      ANALYSERS.each { |analyser| analyser.new(analysed_modules).run }
+      ANALYSERS.each do |analyser_class|
+        analyser_instance = analyser_class.new(analysed_modules)
+        puts "running #{analyser_instance}"
+        analyser_instance.run
+      end
       analysed_modules
     end
 
--- a/lib/rubycritic/analysers/attributes.rb
+++ b/lib/rubycritic/analysers/attributes.rb
@@ -1,10 +1,12 @@
 require "rubycritic/analysers/helpers/methods_counter"
 require "rubycritic/analysers/helpers/modules_locator"
+require "rubycritic/colorize"
 
 module Rubycritic
   module Analyser
 
     class Attributes
+      include Colorize
       def initialize(analysed_modules)
         @analysed_modules = analysed_modules
       end
@@ -13,9 +15,14 @@ module Rubycritic
         @analysed_modules.each do |analysed_module|
           analysed_module.methods_count = MethodsCounter.new(analysed_module).count
           analysed_module.name = ModulesLocator.new(analysed_module).first_name
+          print green "."
         end
+        puts ""
+      end
+
+      def to_s
+        "attributes"
       end
     end
-
   end
 end
--- a/lib/rubycritic/analysers/churn.rb
+++ b/lib/rubycritic/analysers/churn.rb
@@ -1,7 +1,10 @@
+require "rubycritic/colorize"
+
 module Rubycritic
   module Analyser
 
     class Churn
+      include Colorize
       attr_writer :source_control_system
 
       def initialize(analysed_modules)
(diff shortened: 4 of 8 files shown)


guilhermesimoes:
Nice, I really like this approach. And I like how it looks. And it's simple enough that we don't need any more dependencies.

Regarding analyzing file by file instead of analyzer by analyzer. In the past, I've thought about refactoring the code to do just that  but there's at least one analyzer where that wouldn't work: the Flay analyzer, which detects code duplication. For that analyzer to work, it has to parse all files first and only then can it outputs its results.

Is the `red` method in the `Colorize` class supposed to be used when something goes wrong? I dig it.

natesholland:
Yes the idea was that if we ever wanted to indicate that something went wrong we could print a red dot. I don't really know when this would be meaningful but I thought it would be nice to have down the road.

I'll look over this and think about how to clean it up. Are there any immediate things that you think should be tweaked?

guilhermesimoes:
Thinking better, I'll make those tweaks myself, no worries. I'll get this merged this weekend, for sure.

And then I'll finally look into the project rating, sorry for taking so long...

hunterboerner:
Any update on this?

lib/rubycritic/analysers_runner.rb
```diff
@@ -24,7 +24,11 @@ def initialize(paths)
     end
 
     def run
-      ANALYSERS.each { |analyser| analyser.new(analysed_modules).run }
+      ANALYSERS.each do |analyser_class|
+        analyser_instance = analyser_class.new(analysed_modules)
+        puts "running #{analyser_instance}"
```
troessner: Typo, should be "Running"

lib/rubycritic/analysers/smells/flog.rb
```diff
@@ -1,10 +1,12 @@
 require "rubycritic/analysers/helpers/flog"
 require "rubycritic/core/smell"
+require "rubycritic/colorize"
 
 module Rubycritic
   module Analyser
 
     class FlogSmells
+      include Colorize
```
troessner: How about creating something like a BaseAnalyzer that includes Colorize and then have all other analyzers inherit from it?

lib/rubycritic/analysers/smells/flog.rb
```diff
@@ -16,7 +18,13 @@ def initialize(analysed_modules)
       def run
         @analysed_modules.each do |analysed_module|
           add_smells_to(analysed_module)
+          print green "."
```
troessner: This should also not be repeated in every "run" method, we'll just need to figure out a smart way not to so ;)

lib/rubycritic/analysers/smells/flog.rb
```diff
@@ -16,7 +18,13 @@ def initialize(analysed_modules)
...
         end
+        puts ""
+      end
+
+      def to_s
+        "flog smells"
```
troessner: How about doing 

>> self.class.name

instead and then put this method into the BaseAnalyzer (see above)?

troessner:
I'm taking this one over so we can get it merged.

The way I see it 4 things are missing:

1.) Make travis happy
2.) Remove all duplication
3.) Address the comments above
4.) Possibly re-squash to one commit after the changes

Let's get this going ;)

lib/rubycritic/analysers/smells/flog.rb
Onumis: Same goes with the `puts ""`

Onumis:
Completely agree on the `BaseAnalyzer`.

I think that if we go for green dots, we also need red dots, indicating whether smells were detected.
But maybe for now we can just print "uncolored" to indicate that a file was processed.

lucasmazza:
I wonder if the progress indication should be on the reporters implementation or somewhere under the `Rubycritic::Cli` module - the analysers could be just collecting the smells and the CLI feedback is handled elsewhere.

The CodeClimate CLI has an interesting API for something like this: see the [runner](https://github.com/codeclimate/codeclimate/blob/b4a06ece4235c9b7376ee6a317ec5c41a97216e8/lib/cc/analyzer/engines_runner.rb#L21-L25) and [formatter](https://github.com/codeclimate/codeclimate/blob/b4a06ece4235c9b7376ee6a317ec5c41a97216e8/lib/cc/analyzer/formatters/plain_text_formatter.rb#L9-L11) classes to see how it goes.

troessner:
>> I wonder if the progress indication should be on the reporters implementation or somewhere under the Rubycritic::Cli module - the analysers could be just collecting the smells and the CLI feedback is handled elsewhere.

My gut feeling would yell "CLI" now ;)

troessner:
@natesholland are you still on this? Would love to get this one merged ;)

troessner:
I am taking this branch over since @natesholland is not reacting.

troessner:
@Onumis @lucasmazza implementing the progress bar in a DRY and idiomatic object oriented way will require quite some refactoring work on our analysers. We'd have to streamline the "#run" method across all analysers, shift responsibilities and so on. Something we should do anyway but that's bigger topic.

In order to get this going I'd suggest to:

- merge this PR since the feature itself is incredibly valuable (I'm always annoyed by it's absence on every RubyCritic run)
- this will make travis / master red
- right after merge push a fix for the single rubocop offense that makes travis red right now in order to get everything back to green
- work on the above mentioned refactoring that will allow us to implement this in a clean way (I'd volunteer for that)

What do you think?

lucasmazza:
@troessner sounds great to me - if you want to delegate some of the work I can help out during this holiday break :)

Onumis:
@troessner that's a great plan. I'm also available for some work!

troessner:
Alrighty, to finalize this thread, here's what I did:

* merged this one
* [Pushed to master](https://github.com/whitesmith/rubycritic/commit/cb34aec4ac0eb627ea50d7b0b3dce27133805ce3) to fix our rubocop offenses and to get back to a green build as fast as possible
* Created https://github.com/whitesmith/rubycritic/pull/86/files so we can release this into the wild

So I guess we can close this discussion here, thanks for the initiative @natesholland

Issue #30, opened 2015-02-26
Settings file


davidgomes:
Instead of having settings purely as flags, it would be interesting to have a configuration file, which could be ported and used across several projects.

natesholland:
I agree, I think a `rubycritic.yml` file would be a good candidate for how we handle settings.

bglusman:
In [debt_ceiling](https://github.com/bglusman/debt_ceiling) I use the [Confirguations gem](https://github.com/beatrichartz/configurations), and read in a  ruby file with rails like configuration block...  I think I prefer that to yml, but agree having settings/config is a good idea.

johnkelly:
+1. It would be great to specify in a config file which folders ruby critic should critique.

troessner:
Agree, would be useful. Would somebody like to take a stab at this? ;)

troessner:
The same comment I wrote for #11 applies here:

>> Due to RC's nature this is not really trivial since we aggregate multiple tools and those tools have a different understanding of how you can exclude and include paths. And some might not even offer an "exclude" functionality. 
Nevertheless I can see that this feature would be highly valuable. 

;)

nav16:
I am working on this, using yml (`rubycritic.yml`) file.

westlakem:
@nav16 Any updates on this?  I have a project right now where we patch a gem and it is very picky about how we patch it.  This file is the only thing that's not an 'A' and is keeping me from doing a `-s 100` on our ci build

grantneufeld:
It would be nice if the rubycritic config was basically just an extended version of the Reek config — so we could copy our current Reek configuration into it, and add other configuration options for the other tools (flog, flay, etc.) and for rubycritic itself.

Or maybe provide a rubycritic config directory that can contain config files specific to each tool (e.g., `.rubycritic/.reek`).

In particular, I want to be able to configure skipping some of the checks from Flog and Flay (e.g., `HighComplexity`, `DuplicateCode`) for specific directories (e.g., `spec`), without having to ignore those directories for everything else.

Right now, my rubycritic my report gets filled up with lots of complexity and duplicate code smells from my spec directory, unless I don’t include the spec directory (and miss out on reporting the code smells I do want to see for specs).

Thanks!

piacsek:
I guess this can be closed, since we can use `.reek.yml` as a configuration file.

PR #39, opened 2015-03-16
limit churn by commits within the last year

natesholland:
I think limiting the church calculations by 1 year seems reasonable, and greatly speeds up the calculation time for really old projects. Also I think from a metric perspective it makes sense, churn that is over 1 year old doesn't seem like a valuable thing to pay attention to in my mind.

### DIFF
--- a/lib/rubycritic/source_control_systems/git.rb
+++ b/lib/rubycritic/source_control_systems/git.rb
@@ -13,7 +13,7 @@ module Rubycritic
       end
 
       def revisions_count(path)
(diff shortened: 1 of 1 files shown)


guilhermesimoes:
Seems reasonable to me. The Churn gem (in which I based the Churn Analyser), only takes into account the first 3 months when calculating churn. However, it provides an option to configure that date range. I think this is the same use case and as such should be configurable. Guess I gotta start working on #30...

PR #46, opened 2015-03-26
Require Reek 2.0.2.

eostrom:
RubyCritic crashes (because Reek crashes) when analyzed code contains a call to `super` with parameters and a block. I hoped Reek 2.0.2 would fix it. It turns out it doesn't, but it's a known bug (troessner/reek#404) and it's slated to be fixed in 2.0.3.

This PR makes RubyCritic work with 2.0.2. I'm not sure how useful it is on its own, but it should make an eventual upgrade to 2.0.3 (or some future version) easier.


guilhermesimoes:
Wow, thank you so much for this. I watch the Reek repository and for the last couple of months it seems that all I've seen them do is refactor, refactor, refactor.

I was actually dreading having to upgrade to the new Reek version. I guess it wasn't so hard after all :grinning:

PR #47, opened 2015-03-26
Clarified partial-file changeset reference in CONTRIBUTING.md.

eostrom:
I didn't know what `git -p` meant, but `git add -p` makes it easier to find.


guilhermesimoes:
It's not that `git add -p` makes it easier to find what it is, it's that `git -p` is wrong :laughing: Thanks for spotting that.

Also, given that I had to rebase your commit on top of master to make the history linear, I took the liberty of adding yet another link to that CONTRIBUTING file.

Once again, thanks.

Issue #45, opened 2015-03-25
Upgrade to Reek 2.0.3 which fixes calling super with parameters


eostrom:
RubyCritic started crashing when run on my codebase:

```
/home/vagrant/.rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/gems/reek-1.6.5/lib/reek/smells/nested_iterators.rb:73:in `ignored_iterator?': undefined method `method_name' for (super):#<Class:0x007f189d1c0180> (NoMethodError)
```

Reek 2.0 has what [looks like a fix](https://github.com/troessner/reek/commit/7b9af8c63c3f2ca40d47e378a12398122d46706d): "Make NestedIterator not break when iterator is called on super."

I forked the RubyCritic code and tried updating Reek. The tests now fail for two reasons:

  * `samples/reek/smelly.rb` now has two smells, not one; the new one is IrresponsibleModule.
  * `samples/reek/not_smelly.rb` has a lot of smells, not none.

I think the issue in both cases is that `config.reek` is being ignored. The Reek changelog lists a couple of possibly related items involving the Rake task and CLI options.

I'm going to look into that and hopefully submit a PR, but I thought I would check in with what I have so far.

LeeXGreen:
I think this fix *should* be in Reek 1.6.5 as well (see https://github.com/troessner/reek/issues/379), so I'm not sure why you're seeing this.

eostrom:
Good catch, thanks. I guess I have to actually debug.

eostrom:
It looks like the fix handles `super` but not `super()`, which is necessary if, say, you want to make an `Enumerator` subclass whose initializer takes parameters. I'll pursue that over in the Reek repo.

LeeXGreen:
Thanks for the debug :)

topher6345:
I'm also getting this error from rubycritic via reek 1.6.3
`~/.rvm/gems/ruby-2.1.5/gems/reek-1.6.3/lib/reek/smells/nested_iterators.rb:73:in `ignored_iterator?': undefined method `method_name' for (zsuper):#<Class:0x007f9d15c82418> (NoMethodError)`

the code that triggers the error is this: 

```ruby
....
  def initialize(params)
    super do
      @model = MyClass.create! myclass_params
    end
  end
....
```

eostrom:
That one should be fixed in the current release of RubyCritic, which uses
Reek 1.6.5.

topher6345:
[...]

PR #49, opened 2015-04-15
Bump reek dependency to ~> 2.0.4

huerlisi:
Bump reek minimum version to 2.0.4 to get a fix for Issue #45.

Also use optimistic version dependency as reek seems to properly use
semantic versioning.


guilhermesimoes:
Reek's proper use of semantic versioning only applies to its public API, that is, the command line. RubyCritic programmatically uses Reek's internals in a way they were never meant to be used.

[The Reek developers **are** working on a public API for using Reek programmatically][1], but until that time comes, we'll continue to strictly depend on Reek.

If you update this patch, I'll merge it.

[1]: https://github.com/troessner/reek/issues/422#issue-66615287

huerlisi:
Okay, thanks for the information.

Changed to explicit version and amended the commit.

PR #51, opened 2015-04-16
Initial implementation of a RuboCop analyser

DMA57361:
Having spotted mention of RuboCop on the [roadmap](https://github.com/whitesmith/rubycritic/wiki/Roadmap), I thought I'd have a quick go at implementing a RubyCritic analyser for it. This is a first pass that is not yet ready to be part of the gem, but I'd like some feedback on this before I extend the implementation.

Points for consideration:

Do people want this?

From what I can see, RuboCop doesn't have a public API so it is currently run as a single system call, passing in all relevant paths and reading the results back as JSON. Hopefully this should be acceptable. Maybe because of this, rubocop runs using the local .rubocop.yml file in our projects.

This analyser currently adds any results as additional smells. This can make the code page very noisy. Based on what your roadmap says, indications are that the plan would be to have smells and styles as separate groups. I'm happy to do some work to separate smells & styles - probably adding another array of things to AnalysedModule and then supporting this in the generators - however I don't think all RuboCop cops should be considered style issues. My first reaction was Lint & Metrics cops are smells, while Style cops would be style failures. For the Rails cops I'm leaning towards smells, but not sure.

Some of RuboCops cops overlap with existing analysers - it performs three different types of complexity check, for example - should RubyCritic be disabling these? or maybe do nothing and allow users to do so themselves using a local rubocop.yml?


DMA57361:
Please excuse the false start getting the tests to pass, seems I've not quite got a grip on some of the nuances of building gems for public consumption. Will squash that down later if this is going to be used.

PR #55, opened 2015-05-02
add rake task

pmenglund:
lets you use ruby critic in rake without having to "shell out"


pmenglund:
Now I just need to squash it and it should be good to go

troessner:
Taking over this pull request so we can get this on the road, comments incoming ;)

README.md
```diff
@@ -100,6 +100,20 @@ $ rubycritic --help
 | `--deduplicate-symlinks` | De-duplicate symlinks based on their final target     |
 | `--suppress-ratings`     | Suppress letter ratings                               |
 
+
+Rake Task
+---------
+If you want to integrate `rubycritic` into your `Rakefile` just include the below snippet
```
troessner: I believe the right name is "RubyCritic", let's fix that here and below.

lib/rubycritic/rake_task.rb
```diff
@@ -0,0 +1,57 @@
...
+module Rubycritic
+  class RakeTask < Rake::TaskLib
+    attr_accessor :name
+    attr_accessor :report
+    attr_accessor :format
+    attr_accessor :paths
+    attr_accessor :options
+
+    def initialize(*args, &task_block)
+      setup_ivars(args)
```
troessner: "setup_ivars" is not the best name since it ties exactly what it does into the method name. Also I feel this indirection is pointless since it just "makes you look somewhere else" without adding real value.  How about moving this into the initializer?

lib/rubycritic/rake_task.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      desc "Run rubycritic" unless ::Rake.application.last_comment
+
+      task(name, *args) do |_, task_args|
+        RakeFileUtils.send(:verbose, verbose) do
```
troessner: Why do we need this?

lib/rubycritic/rake_task.rb
```diff
@@ -0,0 +1,57 @@
...
+          if task_block
+            task_block.call(*[self, task_args].slice(0, task_block.arity))
+          end
+          rubycritic
```
troessner: How about renaming this method so it is more expressive? It should be a verb in my mind. What do you think of "run"?

lib/rubycritic/rake_task.rb
```diff
@@ -0,0 +1,57 @@
...
+    def initialize(*args, &task_block)
+      setup_ivars(args)
+
+      desc "Run rubycritic" unless ::Rake.application.last_comment
+
+      task(name, *args) do |_, task_args|
+        RakeFileUtils.send(:verbose, verbose) do
+          if task_block
+            task_block.call(*[self, task_args].slice(0, task_block.arity))
+          end
```
troessner: This if block seems overly complex. Maybe it's just something I don't see since I'm kind of new here but could you please explain the rationale behind this?

Ch4s3:
: +1:

rubycritic.gemspec
```diff
@@ -25,9 +25,9 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "flog", "4.2.1"
   spec.add_runtime_dependency "reek", "2.0.4"
   spec.add_runtime_dependency "parser", ">= 2.2.0", "< 3.0"
+  spec.add_runtime_dependency "rake", "~> 10"
```
lucasmazza: I believe that `rake` should be an optional dependency here as developers who are using the RubyCritic executable and not the rake task don't have to install `rake` by default.

README.md
pmenglund: This spells out the command line you use, I can change `rubycritic` to RubyCritic

rubycritic.gemspec
pmenglund: How do you suggest that optional dependency should be implemented? Just leaving it out and catching the load error in `rake_task.rb`?

lucasmazza: I wouldn't even catch the load error - developers requiring the `rake_task` should know upfront that `rake` needs to be installed on their end and the `LoadError` itself is enough feedback on the missing dependency.

pmenglund: ok, an implicit optional dependency then :smile:

lib/rubycritic/rake_task.rb
pmenglund: It makes `rake` output all file operations it performs, but it isn't really needed here as there are none...

lib/rubycritic/rake_task.rb
pmenglund: This lets you call

```ruby
Rubycritic::RakeTask.new(:mytask, [:version] => :test) do |task|
  task.verbose = true
end
```

See https://github.com/ruby/rake/blob/master/lib/rake/dsl_definition.rb#L28-L67

troessner: Hmm, I see. Seems like something where I'm not sure if we really need it. What do you think about dropping this whole RakeFileUtils block?

pmenglund: Dropping the `RakeFileUtils` block is fine, but you need the rest or you won't be able to set stuff like:

```ruby
Rubycritic::RakeTask.new do |task|
  task.paths = "lib"
end
```

lib/rubycritic/rake_task.rb
```diff
@@ -0,0 +1,49 @@
...
+      @name = args.shift || :rubycritic
+      @fail_on_error = true
+      @options = []
+      @format = "html"
+      @report = nil
+      @paths = []
+
+      desc "Run RubyCritic" unless ::Rake.application.last_comment
+
+      task(name, *args) do |_, task_args|
```
troessner: I still dont think we need this (and shouldnt have it since its obscure). 
We dont do this in Reek as well. Check it out, we have an extremely "easy to understand and reason about" definition here: https://github.com/troessner/reek/blob/master/lib/reek/rake/task.rb which allows us to use Reek similar like you intend to do with RubyCritic: https://github.com/troessner/reek/blob/master/tasks/reek.rake

How about setting this up in a similar way here and drop the if branch completely?

README.md
troessner: Yes, lets change it ;)

troessner:
Taking this branch over since @pmenglund is not reacting.

pmenglund:
Sorry! I started on the fix for the feedback, but got interrupted by a higher priority thread...

pmenglund:
Gah! That was not what I intended :tired_face:

pmenglund:
abandoning this and continuing in #85

PR #66, opened 2015-10-18
Add emacs report

pedrocarmona:
Adds emacs report, and allows usage of rubycritic from sublime lint ( https://github.com/pedrocarmona/SublimeLinter-contrib-rubycritic )

### DIFF
--- a/lib/rubycritic/core/location.rb
+++ b/lib/rubycritic/core/location.rb
@@ -18,6 +18,10 @@ module Rubycritic
       "#{pathname}:#{line}"
     end
 
+    def to_s_with_realpath
+      "#{pathname.realpath.to_s}:#{line}"
+    end
+
     def to_h
       {
         :path => pathname.to_s,
--- a/lib/rubycritic/core/location.rb
+++ b/lib/rubycritic/core/location.rb
@@ -19,7 +19,7 @@ module Rubycritic
     end
 
     def to_s_with_realpath
-      "#{pathname.realpath.to_s}:#{line}"
+      "#{pathname.realpath}:#{line}"
     end
 
     def to_h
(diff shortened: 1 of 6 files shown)


lib/rubycritic/core/location.rb
```diff
@@ -18,6 +18,10 @@ def to_s
       "#{pathname}:#{line}"
     end
 
+    def to_s_with_realpath
```
lucasmazza: Since this logic only exists for the emacs renderer it could be just a private method on the `RenderTool` class - what do you think?

lib/rubycritic/generators/emacs/render_tool.rb
```diff
@@ -0,0 +1,34 @@
+require "json"
+
+module Rubycritic
+  module Generator
+    module Emacs
+
+      class RenderTool
```
lucasmazza: Maybe `Renderer` or `Formatter` instead of `RenderTool`?

lucasmazza:
Besides the minor comments this one looks good to me (although I'm not an emacs/sublime user to try it out locally).

lib/rubycritic/core/location.rb
Onumis: Agreed.

lib/rubycritic/generators/emacs/render_tool.rb
Onumis: `Renderer` seems appropriate.

troessner:
@pedrocarmona are you still on this? If not we should close this PR since nobody in the RubyCritic core team uses emacs and thus can take this over.

Onumis:
I was actually thinking on grabbing this one if @pedrocarmona won't, due to the part that:
> allows usage of rubycritic from sublime lint

I think that's neat and Sublime Text is a fairly popular editor nowadays.

troessner:
>> I was actually thinking on grabbing this one

Awesome, I'd then tackle #55 and #36 ;)

Onumis:
Tried to pick this up, but I was not even able to make the linter to work with SublimeText 3, which was the primary use case.

Since this is just a nice-to-have, and I don't want to spend too much time on this, I'm closing it.

troessner:
Zero open pull requests! We have come a long way my friends ;)

Issue #58, opened 2015-07-04
reek 3 has been released


troessner:
reek 3 has been released: https://troessner.wordpress.com/2015/07/04/reek-3-has-been-released/
and one of the biggest feature is a stable, simple API: https://github.com/troessner/reek/blob/master/docs/API.md
Since rubycritic uses a pretty old version of reek (2.0.4 with the current being 3.0.3) I think it would be great if you could upgrade to the latest version.
Not to mention that we fixed quite a few annoying bugs.:)

nijikon:
Can we at least create a new version that has the reek 2.x dependency?

nijikon:
ping

troessner:
@nijikon I guess you have to look into this yourself I'm afraid :(
If you come up with a preliminary pull request pointing at this repo I'd try to help you where I can though ;)

troessner:
We're on 3.6 now, closing this.

nijikon:
@troessner awesome, just updated to 2.1.0 :heart:

PR #67, opened 2015-11-09
Update flay, flog and reek dependencies

lucasmazza:
We are using rubycritic as part of out automatic review process @ Plataformatec and we have been using this branch with updated dependencies (mostly for `reek` latest changes) for a while and has been running smoothly so far. Anyone looking for an update version of rubycritic can check out this fork and try it locally.

**Update** Reek 3 doesn't support Ruby 1.9, so we have a failed build here. Should that be a no-go for using Reek 3 with RubyCritic here?

Related to #58.

### DIFF
--- a/lib/rubycritic/analysers/helpers/reek.rb
+++ b/lib/rubycritic/analysers/helpers/reek.rb
@@ -7,11 +7,9 @@ module Rubycritic
   class Reek < ::Reek::Examiner
     DEFAULT_CONFIG_FILE = File.expand_path("../config.reek", __FILE__)
 
-    def initialize(paths)
-      config = OpenStruct.new(:config_file => DEFAULT_CONFIG_FILE)
-      ::Reek::Configuration::AppConfiguration.initialize_with(config)
-      super(Array(paths))
+    def initialize(pathname)
+      config = ::Reek::Configuration::AppConfiguration.from_path(DEFAULT_CONFIG_FILE)
+      super(pathname, [], :configuration => config)
     end
   end
-
 end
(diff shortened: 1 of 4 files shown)


troessner:
>> Update Reek 3 doesn't support Ruby 1.9, so we have a failed build here. Should that be a no-go for using Reek 3 with RubyCritic here?

It was a conscious decision to remove Ruby 1.9 support from "Reek" back then since 1.9 is not supported anymore. So I'd rather suggest the other way around and drop 1.9 support in rubycritic as well.

pboling:
Using the `gemfiles` pattern and build matrix exclusions on Travis it should not be too hard to support either reek 2 or 3, I think.

lucasmazza:
> So I'd rather suggest the other way around and drop 1.9 support in rubycritic as well.

I'm :+1: on that too, and I wouldn't go as far as attempt to support different versions of Reek since this is mostly a standalone CLI program rather than an additional dependency that users can cherrypick which Reek version they want to use.

pboling:
That's a big decision though, would be a major version bump, right?

troessner:
Yep,,looks like a major bump to me as well ;)

guilhermesimoes:
Reek is also failing to install in JRuby. I can see that Reek is allowing failures in JRuby, so maybe RubyCritic's travis config should be updated later on.

Anyway, I'll make an announcement later on for you guys. Thanks for sticking with RubyCritic for so long.

troessner:
PR itself looks good to me, let's clarify #68 and then get this bad boy merged ;)

troessner:
@lucasmazza could you please rebase your PR against master? With the merging of #69 this should now be green on travis.:)

lucasmazza:
@troessner rebased :+1:

troessner:
Merged, thanks a lot @lucasmazza !

PR #69, opened 2015-11-10
Drop support for Ruby 1.9.

### LINKED ISSUE #68: Drop support for 1.9
In the light of #67 I suggest to drop support for 1.9 completely.
Support for ruby 1.9 has been dropped for quite some while, we should let it die here as well.

From what I can see in order to make this happen we'd need to:

* remove it from travis. We could also make it an allowed failure but I don't see the point in this specific case
* major version bump 
* explicitly list it on our README

What do you guys think?

troessner:
Fixes #68 
I'd vote for then having a separate "bump version to 2.0" pull request and then release 2.0 as fast as possible so we can get #67 going ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,17 @@
+sudo: false
+cache: bundler
 language: ruby
-
+script:
+  - bundle exec rake
 rvm:
-  - 1.9.3
-  - 2.0.0
+  - 2.0
   - 2.1
   - 2.2
-  - jruby
   - rbx-2
+  - jruby
+matrix:
+  allow_failures:
+    - rvm: rbx-2
+    - rvm: jruby
+  fast_finish: true
 
-script: "bundle exec rake"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,15 @@
-sudo: false
-cache: bundler
 language: ruby
-script:
-  - bundle exec rake
+
 rvm:
   - 2.0
   - 2.1
   - 2.2
-  - rbx-2
   - jruby
+  - rbx-2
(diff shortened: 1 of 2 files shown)


pboling:
I'd recommend creating a permanent, stable, branch for the current major version, before merging in stuff for the next major version, as a reference point for people stuck in the old world.

guilhermesimoes:
Why is RuboCop running inside `/home/travis/build/whitesmith/rubycritic/vendor/bundle/ruby/2.0.0/gems/`? The build is failing this way.

troessner:
>> I'd recommend creating a permanent, stable, branch for the current major version, before merging in stuff for the next major version, as a reference point for people stuck in the old world.

We literally had the same discussion [https://github.com/troessner/reek/issues/401](in Reek) and decided against it. The thing is, IF you really need support for 1.9 you just....have to stick with the last major version that supports it. I'll add a prominent note in the README so people can easily spot it.
And dropping 1.9 is not only a travis thing: as soon as this is merged we can and should use all the niceties that ruby 2 brings with it. In "Reek" this was mostly used in the form of keyword arguments that cleaned up our code base significantly. And as soon as you introduce this a stable branch is useless anyway because you can't backport anything without tremendous effort.

>> Why is RuboCop running inside /home/travis/build/whitesmith/rubycritic/vendor/bundle/ruby/2.0.0/gems/? The build is failing this way.

I'll look into this.

Onumis:
I think we just create a tag for the latest known version that supports ruby 1.9 and mention it on the README like @troessner suggested.

Good work on fixing travis.

PR #72, opened 2015-11-11
Bump version to 2.1.

troessner:
We merged #67 so let's get a new version out there.
The dependency bump is significant, 2 + 1 minor version bumps for flay and flog and a major version bump for Reek. This warrants a major version bump in RubyCritic as well in my mind.


Onumis:
:+1:

PR #85, opened 2015-12-18
add rake task

pmenglund:
starting over from #55 as it got botched

### DIFF
--- a/lib/rubycritic/rake_task.rb
+++ b/lib/rubycritic/rake_task.rb
@@ -1,0 +1,50 @@
+require "rake"
+require "rake/tasklib"
+
+module Rubycritic
+  class RakeTask < ::Rake::TaskLib
+    attr_accessor :name
+    attr_accessor :report
+    attr_accessor :format
+    attr_accessor :paths
+    attr_accessor :options
+
+    def initialize(*args, &task_block)
+      @name = args.shift || :rubycritic
+      @fail_on_error = true
+      @options = []
+      @format = "html"
+      @report = nil
+      @paths = []
+
+      desc "Run RubyCritic" unless ::Rake.application.last_comment
+
+      task(name, *args) do |_, task_args|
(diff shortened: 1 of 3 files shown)


lib/rubycritic/rake_task.rb
```diff
@@ -0,0 +1,50 @@
...
+      @name = args.shift || :rubycritic
+      @fail_on_error = true
+      @options = []
+      @format = "html"
+      @report = nil
+      @paths = []
+
+      desc "Run RubyCritic" unless ::Rake.application.last_comment
+
+      task(name, *args) do |_, task_args|
```
troessner: You kind of completely ignored my comment from your old pull request :-1: 
So let me repeat:

I still dont think we need this (and shouldnt have it since its obscure). 
We dont do this in Reek as well. Check it out, we have an extremely "easy to understand and reason about" definition here: https://github.com/troessner/reek/blob/master/lib/reek/rake/task.rb which allows us to use Reek similar like you intend to do with RubyCritic: https://github.com/troessner/reek/blob/master/tasks/reek.rake

How about setting this up in a similar way here and drop the if branch completely?

troessner:
I see 3 obstacles before we can merge this:

1.) Travis fails
2.) Let's add some spec coverage 
3.) Please see my comments

Other than that, nice job, let's get this going!

lib/rubycritic/rake_task.rb
pmenglund: Bugger! I had changed that, but when I messed up the rebase, I apparently managed to merge the wrong thing. I had copied what you do in reek...

I'll fix it!

troessner:
.......another 2 weeks have passed :(
@pmenglund can you please push your changes?

troessner:
Closing this one. Sorry, but those huge delays in back and forth just don't work for me. That's too much friction and too time consuming.

I'll look into implementing the rake task later.

PR #88, opened 2015-12-29
prepend gem lib path when using rubycritic executable

jbodah:
This makes it a little easier to do dev on the gem locally by allowing you to call `bin/rubycritic` which will look in your dev copy instead of pulling an installed gem's files

### DIFF
--- a/bin/rubycritic
+++ b/bin/rubycritic
@@ -1,5 +1,9 @@
 #!/usr/bin/env ruby
 
+# Always look in the lib directory of this gem
+# first when searching the load path
+$LOAD_PATH.unshift File.expand_path("../../lib", __FILE__)
+
 require "rubycritic/cli/application"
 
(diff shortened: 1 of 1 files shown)


bin/rubycritic
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env ruby
 
+# Always look in the lib directory of this gem
+# first when searching the load path
+$:.unshift File.expand_path("../../lib", __FILE__)
```
troessner: I really can't wait for Ruby's shitty shellism's to die.
Let's replace 

```$:```

 with

```$LOAD_PATH```

troessner:
Good idea! One small change (please see my comment) and then let's get this merged!

jbodah:
Cool fixed

troessner:
Merged, thanks for the contribution!

Onumis:
Love it!

Issue #89, opened 2015-12-30
add a console report


jbodah:
AFAIK the only outputs we have are json file and html file. It would be nice to just enable it to print a summary to the screen instead of to a file

Might take a go at this some time soon, just want to document it

troessner:
Yes, that would be really nice, I agree.

>> Might take a go at this some time soon

By all means, please go ahead ;)

troessner:
Closing this, since #91 has been merged.

PR #91, opened 2015-12-30
add a ConsoleReport format

jbodah:
addresses https://github.com/whitesmith/rubycritic/issues/89

1__thanks_for_flying_vim__tmux_


lib/rubycritic/core/analysed_module.rb
```diff
@@ -43,6 +43,10 @@ def smells_at_location(location)
       smells.select { |smell| smell.at_location?(location) }
     end
 
+    def <=>(other)
+      [rating.to_s, name] <=> [other.rating.to_s, other.name]
```
jbodah: Maybe we want this to sort bad ratings first?

lib/rubycritic/core/rating.rb
```diff
@@ -22,6 +22,16 @@ def to_h
       @letter
     end
 
+    def color
+      case @letter
+      when "A" then :green
```
jbodah: Symbols based on colorize; could refactor the html reporter to switch based on the symbol if we wanted

lib/rubycritic/generators/console_report.rb
```diff
@@ -0,0 +1,52 @@
...
+          report << "\n"
+          report << smell_reports.join("\n")
+        end
+        report
+      end
+
+      def generate_smell_report(smell)
+        smell_line = smell.to_s
+        report = <<-EOF
+     * #{smell_line}
```
jbodah: Not sure if there's a preferred alternative to add the indentation

lib/rubycritic/cli/options.rb
```diff
@@ -19,10 +19,11 @@ def parse
 
           opts.on(
             "-f", "--format [FORMAT]",
-            [:html, :json],
+            [:html, :json, :console],
```
jbodah: Might be nice to pull these dynamically. Left this way for simplicity

lib/rubycritic/core/rating.rb
```diff
@@ -22,6 +22,16 @@ def to_h
...
+      when "B" then :green
+      when "C" then :yellow
+      when "D" then :light_red
+      when "F" then :red
+      end
```
troessner: Let's rather use a constant pointing to a hash like this:

```Ruby
RatingToColor = {
  A: :green,
  # ....
}
```

This will express our intent clearer (after all, we're talking about a constant mapping), will provide a clear separation of code vs data (in the sense: there is no code since it's just data) and be more performant for the same reasons.

lib/rubycritic/generators/console_report.rb
```diff
@@ -0,0 +1,52 @@
...
+        reports = @analysed_modules.sort.map do |analysed_module|
+          generate_module_report analysed_module
+        end
+        puts reports.join("\n")
+      end
+
+      private
+
+      def generate_module_report(mod)
+        report = <<-EOF
```
troessner: I know why you're doing this but I still find it very detrimental for readilbility. I'd rather use something like [here](http://rubyquicktips.com/post/4438542511/heredoc-and-indent)

lib/rubycritic/generators/console_report.rb
```diff
@@ -0,0 +1,52 @@
...
+#{mod.name.send(mod.rating.color)}:
```
troessner: [Law of Demeter](http://devblog.avdi.org/2011/07/05/demeter-its-not-just-a-good-idea-its-the-law/) violation

lib/rubycritic/generators/console_report.rb
```diff
@@ -0,0 +1,52 @@
...
+    Rating:       #{mod.rating.to_s.send(mod.rating.color)}
```
troessner: Same as above. Apart from that it's hard to read and hard to maintain. We should refactor our approach of how to access those attributes using a separate report object so that we can use proper composition and delegation.

lib/rubycritic/generators/console_report.rb
```diff
@@ -0,0 +1,52 @@
...
+        end
+        if smell_reports.any?
+          report << "\n"
+          report << smell_reports.join("\n")
+        end
+        report
+      end
+
+      def generate_smell_report(smell)
+        smell_line = smell.to_s
```
troessner: "smell.to_s" returns "smell_line"? That's pretty obscure.

lib/rubycritic/generators/console_report.rb
```diff
@@ -0,0 +1,52 @@
...
+      def generate_report
+        reports = @analysed_modules.sort.map do |analysed_module|
+          generate_module_report analysed_module
+        end
+        puts reports.join("\n")
+      end
+
+      private
+
+      def generate_module_report(mod)
```
troessner: SRP violation / FeatureEnvy - Doesn't a method like this actually belong to an "analysed_module"?

lib/rubycritic/generators/console_report.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      def generate_smell_report(smell)
+        smell_line = smell.to_s
+        report = <<-EOF
+     * #{smell_line}
+        EOF
+        smell.locations.each do |location|
+          report << <<-EOF
+       - #{location}
+          EOF
```
troessner: All those multiple heredocs make this method very hard to read. Additionally, I believe this method does way to much - we should split this up into small, maintainable methods.

lib/rubycritic/generators/console_report.rb
troessner: Errrrm, and you're calling `send` on `to_s`. That feels wrong on quite a few levels ;)

troessner:
Love the idea! 
Regarding the implementation I do feel like there's still some work left to do ;)

lib/rubycritic/generators/console_report.rb
jbodah: How do you feel about using ERB here instead?

lib/rubycritic/generators/console_report.rb
jbodah: This is to utilize colorize. The only other thing that comes to mind is to split out each rating into a separate class and have a `#colored` or something. Do you have a different suggestion?

lib/rubycritic/generators/console_report.rb
jbodah: Woops, old code

lib/rubycritic/generators/console_report.rb
```diff
@@ -0,0 +1,22 @@
...
+      TEMPLATE_PATH = File.expand_path("../../generators/text/templates/analysed_module_console_report.erb", __FILE__)
+
+      def initialize(analysed_modules)
+        @analysed_modules = analysed_modules
+      end
+
+      def generate_report
+        template = ERB.new(File.read(TEMPLATE_PATH), nil, '%<>-')
+        reports = @analysed_modules.sort.map do |mod|
+          template.result(mod.instance_eval { binding })
```
jbodah: Kind of sucks, maybe there's a better approach

lib/rubycritic/core/rating.rb
```diff
@@ -1,6 +1,16 @@
+require "colorize"
+
 module Rubycritic
 
   class Rating
+    RATING_TO_COLOR = {
```
lucasmazza: This colouring logic should be on `ConsoleReport`, not here - a private method like `color_for_rating` or something like that and pass both `mod` and the color to the template :)

lib/rubycritic/generators/text/templates/analysed_module_console_report.erb
```diff
@@ -0,0 +1,12 @@
+<%= name.send(rating.color) %>:
```
lucasmazza: let's avoid `send` and use the `colorize` method to be more explicit about the coloring usage, wdyt?

lucasmazza:
> Love the idea!

:+1: loved the output too

lib/rubycritic/generators/text/templates/analysed_module_console_report.erb
troessner: And **if** we ever use `send` it should be `public_send` ;)

lib/rubycritic/core/rating.rb
troessner: :+1: !

troessner:
Ahhhhhhh, this looks a lot nicer already :)
Let's address the 2 remaining comments from @lucasmazza and then I think this little rascal is ready for merge!

lib/rubycritic/generators/text/templates/analysed_module_console_report.erb
jbodah: Ah I missed that method, thanks :smile_cat:

rubycritic.gemspec
```diff
@@ -25,6 +25,8 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "flog", "4.3.2"
   spec.add_runtime_dependency "reek", "3.7.1"
   spec.add_runtime_dependency "parser", ">= 2.2.0", "< 3.0"
+  spec.add_runtime_dependency "colorize"
+  spec.add_runtime_dependency "tilt"
```
jbodah: added Tilt if we're okay with that

lucasmazza: I think we can implement the template scoping without Tilt by implementing some `Template`-ish class:

```ruby
Template = Struct.new(:analyzed_module, :color) do
  def render
    erb = ERB.new # ...
    erb.render(binding)
  end
end

reports = @analysed_modules.sort.map do |mod|
  Template.new(mod, RATING_TO_COLOR[mod.rating.to_s]).render
end
```

With this the ERB template will have access to a `analyzed_module` and `color` methods. What do you think?

troessner: @jbodah can you please address this last comment so we can get this bad boy merged? 
Can't wait to see this live :)

jbodah:
Addressed feedback

lib/rubycritic/generators/text/list.rb
```diff
@@ -0,0 +1,43 @@
...
+
+module Rubycritic
+  module Generator
+    module Text
+      class List
+        class << self
+          TEMPLATE_PATH = File.expand_path("../templates/list.erb", __FILE__)
+
+          def erb_template
+            @erb_template ||= ERB.new(File.read(TEMPLATE_PATH), nil, '-')
```
lucasmazza: Can you please use double quotes on this `-` String to make the RuboCop checks green?

jbodah: Fixed

lucasmazza:
Awesome :metal: @troessner anything else?

troessner:
Very nice @jbodah ! From a code base POV we could merge this bad boy **but** there is still one biggie missing. No specs. Seriously? :)
Let's at least spec `ConsoleReport#generate_report` and then I'm happy to merge this one!

Onumis:
This looks really good, great job @jbodah!

jbodah:
Will get some specs out for this within the next few days, a bit swamped with everyone getting back from the holidays

jbodah:
added test for `ConsoleReport#generate_report`

troessner:
Well, the test coverage is still spotty since you only test that something is written to stdout, but never check what exactly :)
But in order to not make this the never-ending story I'm willing to let this go now.
Please squash to one commit so we can merge this one.

troessner:
@jbodah i'd love to get this merged - coulx you please squash to one commit?

jbodah:
squashed

troessner:
Awesome! ButI am afraid you need to make travis happy as well ;)

jbodah:
> Awesome! ButI am afraid you need to make travis happy as well ;)

No idea what this is about, but I can't reproduce it. Seems totally unrelated given the only gem change was colorize which doesn't rely on anything:

```
Bundler could not find compatible versions for gem "parser":

  In Gemfile:

    rubycritic (>= 0) ruby depends on

      reek (= 3.8.1) ruby depends on

        parser (>= 2.2.2.5, ~> 2.2) ruby

    rubycritic (>= 0) ruby depends on

      reek (= 3.8.1) ruby depends on

        unparser (~> 0.2.2) ruby depends on

          parser (2.2.0.2)
```

jbodah:
> I would expect this to be failing on master as well

Reproduced with a dummy commit: https://travis-ci.org/whitesmith/rubycritic/builds/102826574

jbodah:
Apparently it has *something* to do with bundler since updating that did... *something*

jbodah:
related: https://github.com/travis-ci/travis-ci/issues/3531

troessner:
Merging this bad boy!
Nice job, thanks!

troessner:
New release incomimg ..;)

PR #92, opened 2015-12-30
delegate Rubycritic.create to CommandFactory

jbodah:
Small suggestion. I kept forgetting that `Rubycritic.create` returned a `Command` instance

### DIFF
--- a/lib/rubycritic.rb
+++ b/lib/rubycritic.rb
@@ -1,22 +1,7 @@
-require "rubycritic/configuration"
+require "rubycritic/command_factory"
 
 module Rubycritic
   def self.create(options = {})
-    options_hash = options.to_h
-    Config.set(options_hash)
-    case Config.mode
-    when :version
-      require "rubycritic/commands/version"
-      Command::Version.new
-    when :help
-      require "rubycritic/commands/help"
-      Command::Help.new(options.help_text)
-    when :ci
-      require "rubycritic/commands/ci"
-      Command::Ci.new(options_hash[:paths])
-    else
-      require "rubycritic/commands/default"
-      Command::Default.new(options_hash[:paths])
-    end
+    CommandFactory.create(options)
   end
 end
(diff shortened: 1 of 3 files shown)


lib/rubycritic.rb
```diff
@@ -1,22 +1,7 @@
-require "rubycritic/configuration"
+require "rubycritic/command_factory"
 
 module Rubycritic
   def self.create(options = {})
```
troessner: Why not go all the way and remove RubyCritic.create so that callers have to use RubyCritic::CommandFactory.create? 
That would be even more expressive.

jbodah: Didn't want to break any public API's. I'm fine with that if you are

troessner: AFAIK RubyCritic never had a communicated, agreed on and versioned API. So this should be fine.
@guilhermesimoes what do you think about this?

Personally I have the guideline that unless a project has  a separate piece of documentation that says "API" then it doesn't have one I can rely on. This keeps my expectation in check.

guilhermesimoes: @troessner I agree with that guideline. I say go for it!

Informally I presented @bglusman with a way to use RubyCritic programatically (in his deb_ceiling gem) but I dare to say that compatibility was never a first priority...

Given that you guys are in a refactoring process and that there seems to be some community interest in using RubyCritic programatically... May I suggest you agree on an API so that it can be documented and supported? I would love to see that :smile:

lucasmazza: :+1: on that. This part of Rubycritic always felt more an CLI abstraction layer rather than a programatically API, so I think we should refactor it as we want it - The `AnalysersRunner` sounds more like the way to go if someone wants an API (which is what we do in an internal app :smile:)

troessner: Created #93 out of this ;)

troessner:
@jbodah nice job, merging this one!

P.S.: In the future, could you please write a quick comment a la "Addressed all comments" when you pushed? I hadn't noticed that you had addressed all changes already since you get no notifications for pushes on pull requests...

Issue #59, opened 2015-07-23
Incomplete constant name if file has no methods


jbodah:
Given

```rb
module Office365
  # This class represents the cached user information retrieved from the
  # ActiveDirectory Graph API. It is a user that exists within a Microsoft
  # Azure Active Directory. It is used for user management.
  class User < ActiveRecord::Base
  # ...
```

Instead of showing up as `Office365::User` it simply shows up as `User` (which also shows up for `app/models/user.rb`)

<img width="796" alt="rubycritic" src="https://cloud.githubusercontent.com/assets/6093002/8842649/c1d9f532-30ca-11e5-8891-69802870d101.png">

troessner:
Actually I'd consider this a bug. Re-labeled. Let's fix this.

Onumis:
This is currently a feature: [when a file has no methods, returns the name of the file titleized](https://github.com/whitesmith/rubycritic/blob/master/test/lib/rubycritic/analysers/helpers/modules_locator_test.rb#L41)

I don't mind having this changed to include the module name (if it exists) for classes without methods.

The conditions to having the name of the file displayed instead of the "qualified name" are:

- file is empty
- file is unparsable
- file has no methods

troessner:
I'd prefer to have the module name always included as well.

troessner:
#96 has been merged -> closed!

PR #96, opened 2016-01-08
Display the class and module names when the file has no methods.

Onumis:
Proposal fix for #59

### DIFF
--- a/test/lib/rubycritic/core/analysed_modules_collection_test.rb
+++ b/test/lib/rubycritic/core/analysed_modules_collection_test.rb
@@ -26,7 +26,7 @@ describe Rubycritic::AnalysedModulesCollection do
       let(:paths) { "test/samples/" }
 
       it "recursively registers all files" do
-        subject.count.must_equal 13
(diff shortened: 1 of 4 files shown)


test/lib/rubycritic/core/analysed_modules_collection_test.rb
```diff
@@ -26,7 +26,7 @@
       let(:paths) { "test/samples/" }
 
       it "recursively registers all files" do
-        subject.count.must_equal 13
+        subject.count.must_equal 14
```
Onumis: This is a pita... we have to increment this if we add new sample files to test with...

What about we change to: `subject.count.must_equal Dir['test/samples/**/*.rb'].count`?

troessner: Good idea!

troessner:
One comment and then lgtm!

Onumis:
@troessner updated!

troessner:
Nice! Merged ;)

Issue #98, opened 2016-01-12
Break on Continuous Integration based on a threshold


RobertoSchneiders:
We are migrating our code analisys from code climate (which is awesome) to open source tools and we are using rubycritic (also awesome). 

We want to break the build on CI if the quality is bellow a threshold. Unfortunately I can't do this easily with rubycritic. I had to create a pretty crazy shell [script](https://gist.github.com/RobertoSchneiders/f8b319440accf3ad4da5). 

I'm wondering if you guys think this could be implemented inside rubycritic, what do you think?

I found this gem [quality-control](https://github.com/gdotdesign/quality-control) which basicaly do exactly that (they also have rating threshold, which sounds good), but, I'm trying to avoid this dependency. (The project don't seems to be very active)

troessner:
Yes, I would **love** this functionality. 
WDYT about a cli switch like "--minimum-score" based on [our current scoring](https://github.com/whitesmith/rubycritic/pull/35)?

RobertoSchneiders:
@troessner pretty much what I had in mind. :+1:

RobertoSchneiders:
Any insights on how we can do this?

troessner:
[1] Add the cli option to our [options](https://github.com/whitesmith/rubycritic/blob/master/lib/rubycritic/cli/options.rb)
[2] Check the calculated [score](https://github.com/whitesmith/rubycritic/blob/master/lib/rubycritic/core/analysed_modules_collection.rb#L32)
[3] Then change our [generators](https://github.com/whitesmith/rubycritic/blob/master/lib/rubycritic/generators/json/simple.rb)

Implementing this is pretty easy. Doing this "the right away" without making our overall architecture is not.
[1] and [2] is straightforward, [3] not:

* AFAIS right now our reporters / generators have no notion of setting a proper exit status (so they just use the default)
* Additionally to that you'd need to "carry" this configuration setting through in a clean extensible way from beginning to the end (because it's the first thing RubyCritic gets handed and the last thing it needs to evaluate)

Would you like to take a stab at this?
I'd help you with constant reviews on a preliminary pull request ;)

RobertoSchneiders:
Great! Yeah, I will try. I think I have some time on friday.

RobertoSchneiders:
Released on `2.6.0`.

PR #99, opened 2016-01-13
Add a minimum score option to the command line interface

RobertoSchneiders:
This is the first step for #98. It work, but, I'm not completely happy with some choices, maybe you can provide some nice suggestions :)

First of all, I'm assuming that the responsibility for the exit code should be in the cli, not in the lib itself. Also, the minimum score feature belongs to the cli and not the lib itself. So, I think it is plausible to expect the lib commands to return the score. The problem with that is not all commands will have a score (eg. --help), so, it's not a great solution. Anyway, I implemented that way for now. Any suggestions are welcome.

I made a change in the CI command, now it inherit from the default command and overrides the `critique` method (The only difference between them). I made this to remove some code duplication. I don't know if this was the best thing to do, so, a feedback will be appreciated.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,7 @@ AllCops:
     - '**/Rakefile'
   Exclude:
     - 'test/samples/**/*'
+    - 'tmp/**/*'
   # By default, the rails cops are not run. Override in project or home
   # directory .rubocop.yml files, or by giving the -R/--rails option.
   RunRailsCops: false
--- a/features/command_line_interface/minimum_score.feature
+++ b/features/command_line_interface/minimum_score.feature
@@ -1,0 +1,39 @@
+Feature: Break if overall score is bellow minimum
+  In order to break the Continuous Integration builds based on a score threshold
+  Rubycritic returns the exit status according with the score
+
+  Scenario: Status indicates a success when not using --minimum-score
+    Given the smelly file 'smelly.rb' with a score of 93.75
+    When I run rubycritic smelly.rb
+    Then the exit status indicates a success
+
+  Scenario: Status indicates an error when score below the minimum
+    Given the smelly file 'smelly.rb' with a score of 93.75
+    When I run rubycritic --minimum-score 100 smelly.rb
+    Then the exit status indicates an error
+
+  Scenario: Status indicates a success when score is above the minimum
+    Given the smelly file 'smelly.rb' with a score of 93.75
+    When I run rubycritic --minimum-score 93 smelly.rb
+    Then the exit status indicates a success
+
+  Scenario: Status indicates a success when score is equal the minimum
+    Given the clean file 'clean.rb' with a score of 100
+    When I run rubycritic --minimum-score 100 clean.rb
+    Then the exit status indicates a success
+
+  Scenario: Prints the score on output
+    Given the smelly file 'smelly.rb' with a score of 93.75
+    When I run rubycritic smelly.rb
+    Then the output should contain:
+    """
+    Score: 93.75
+    """
+
+  Scenario: Prints a message informing the score is below the minimum
+    Given the empty file 'empty.rb' with a score of 0
+    When I run rubycritic --minimum-score 100 empty.rb
+    Then the output should contain:
+    """
+    Score (0.0) is below the minimum 100
+    """
--- a/features/command_line_interface/options.feature
+++ b/features/command_line_interface/options.feature
@@ -1,0 +1,35 @@
+Feature: Rubycritic can be controlled using command-line options
+  In order to change Rubycritic's default behaviour
+  As a developer
+  I want to supply options on the command line
+
+  Scenario: return non-zero status on bad option
+    When I run rubycritic --no-such-option
+    Then the exit status indicates an error
+    And it reports the error "Error: invalid option: --no-such-option"
+    And there is no output on stdout
+
+  Scenario: display the current version number
+    When I run rubycritic --version
+    Then it succeeds
+    And it reports the current version
+
+  Scenario: display the help information
+    When I run rubycritic --help
+    Then it succeeds
+    And it reports:
+      """
+      Usage: rubycritic [options] [paths]
+          -p, --path [PATH]                Set path where report will be saved (tmp/rubycritic by default)
+          -f, --format [FORMAT]            Report smells in the given format:
+                                             html (default)
+                                             json
+                                             console
+          -s, --minimum-score [MIN_SCORE]  Set a minimum score
+          -m, --mode-ci                    Use CI mode (faster, but only analyses last commit)
+              --deduplicate-symlinks       De-duplicate symlinks based on their final target
+              --suppress-ratings           Suppress letter ratings
+          -v, --version                    Show gem's version
+          -h, --help                       Show this message
+
+      """
--- a/lib/rubycritic/cli/application.rb
+++ b/lib/rubycritic/cli/application.rb
@@ -14,11 +14,17 @@ module Rubycritic
 
       def execute
         parsed_options = @options.parse
-        ::Rubycritic::CommandFactory.create(parsed_options).execute
-        STATUS_SUCCESS
+
+        reporter = Rubycritic::CommandFactory.create(parsed_options.to_h).execute
+        print(reporter.status_message)
+        reporter.status
       rescue OptionParser::InvalidOption => error
         $stderr.puts "Error: #{error}"
         STATUS_ERROR
+      end
+
+      def print(message)
+        $stdout.puts message
       end
     end
   end
--- a/lib/rubycritic/cli/options.rb
+++ b/lib/rubycritic/cli/options.rb
@@ -27,6 +27,10 @@ module Rubycritic
             @format = format
           end
 
+          opts.on("-s", "--minimum-score [MIN_SCORE]", "Set a minimum score") do |min_score|
+            @minimum_score = Integer(min_score)
+          end
+
           opts.on("-m", "--mode-ci", "Use CI mode (faster, but only analyses last commit)") do
             @mode = :ci
           end
@@ -50,10 +54,6 @@ module Rubycritic
         self
       end
 
-      def help_text
-        @parser.help
-      end
-
       def to_h
         {
           :mode => @mode,
@@ -61,7 +61,9 @@ module Rubycritic
           :format => @format,
           :deduplicate_symlinks => @deduplicate_symlinks,
           :paths => paths,
-          :suppress_ratings => @suppress_ratings
+          :suppress_ratings => @suppress_ratings,
+          :help_text => @parser.help,
+          :minimum_score => @minimum_score || 0
         }
       end
 
--- a/lib/rubycritic/command_factory.rb
+++ b/lib/rubycritic/command_factory.rb
@@ -3,21 +3,24 @@ require "rubycritic/configuration"
 module Rubycritic
   class CommandFactory
     def self.create(options = {})
-      options_hash = options.to_h
-      Config.set(options_hash)
-      case Config.mode
+      Config.set(options)
+      command_class(Config.mode).new(options)
+    end
+
+    def self.command_class(mode)
+      case mode
       when :version
         require "rubycritic/commands/version"
-        Command::Version.new
+        Command::Version
       when :help
         require "rubycritic/commands/help"
-        Command::Help.new(options.help_text)
+        Command::Help
       when :ci
         require "rubycritic/commands/ci"
-        Command::Ci.new(options_hash[:paths])
+        Command::Ci
       else
         require "rubycritic/commands/default"
-        Command::Default.new(options_hash[:paths])
+        Command::Default
       end
     end
   end
--- a/lib/rubycritic/commands/base.rb
+++ b/lib/rubycritic/commands/base.rb
@@ -1,0 +1,16 @@
+require "rubycritic/commands/status_reporter"
+
+module Rubycritic
+  module Command
+    class Base
+      def initialize(options)
+        @options = options
+        @status_reporter = Rubycritic::Command::StatusReporter.new(@options)
+      end
+
+      def execute
+        raise NotImplementedError
+      end
+    end
+  end
+end
--- a/lib/rubycritic/commands/status_reporter.rb
+++ b/lib/rubycritic/commands/status_reporter.rb
@@ -1,0 +1,43 @@
+module Rubycritic
+  module Command
+    class StatusReporter
+      attr_reader :status, :status_message
+      SUCCESS = 0
+      SCORE_BELOW_MINIMUM = 1
+
+      def initialize(options)
+        @options = options
+        @status = SUCCESS
+      end
+
+      def score=(score)
+        @score = score
+        update_status
+      end
+
+      private
+
+      def update_status
+        @status = current_status
+        update_status_message
+      end
+
+      def current_status
+        satisfy_minimum_score_rule ? SUCCESS : SCORE_BELOW_MINIMUM
+      end
+
+      def satisfy_minimum_score_rule
+        @score >= @options[:minimum_score]
+      end
+
+      def update_status_message
+        case @status
+        when SUCCESS
+          @status_message = "Score: #{@score}"
+        when SCORE_BELOW_MINIMUM
+          @status_message = "Score (#{@score}) is below the minimum #{@options[:minimum_score]}"
+        end
+      end
+    end
+  end
+end
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ $ rubycritic --help
 |--------------------------|-------------------------------------------------------|
 | `-v/--version`           | Displays the current version and exits                |
 | `-p/--path`              | Sets the output directory (tmp/rubycritic by default) |
+| `-s/--minimum-score`     | Set a minimum score                                   |
 | `--mode-ci`              | Uses CI mode (faster, but only analyses last commit)  |
 | `--deduplicate-symlinks` | De-duplicate symlinks based on their final target     |
 | `--suppress-ratings`     | Suppress letter ratings                               |
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,7 @@
 require "bundler/gem_tasks"
 require "rake/testtask"
 require "rubocop/rake_task"
+require "cucumber/rake/task"
 
 Rake::TestTask.new do |task|
   task.libs.push "lib"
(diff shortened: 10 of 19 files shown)


RobertoSchneiders:
Outputs:
```
$rubycritic lib --minimum-score 98
...
New critique at file:///...
Score (97.875) is below the minimum 98.0

$echo $?
1
```
```
rubycritic lib --minimum-score 97
...
New critique at file:///...
Score: 97.875

$echo $?
0
```

lib/rubycritic/cli/application.rb
```diff
@@ -14,12 +14,29 @@ def initialize(argv)
...
         STATUS_SUCCESS
       rescue OptionParser::InvalidOption => error
         $stderr.puts "Error: #{error}"
         STATUS_ERROR
       end
+
+      private
+
+      def satisfy_minimum_score?(result, options)
+        return true unless result.is_a?(Hash) && result[:score]
```
troessner: Let's not do this. This is 1.) extremely obscure (no innocent reader will understand why you are doing this at first glance) and 2.) makes heavy use of type checking, something that is smelly per se in a dynamically typed language like Ruby.
We should test here that we satisfy the minimum-score, nothing else. The check that this option is set should happen above, in "execute".

lib/rubycritic/cli/application.rb
```diff
@@ -14,12 +14,29 @@ def initialize(argv)
...
+
+        status = result[:score] >= options.minimum_score
+        if status
+          $stdout.puts "Score: #{result[:score]}"
+        else
+          $stderr.puts "Score (#{result[:score]}) is below the minimum #{options.minimum_score}"
+        end
```
troessner: The method is a predicate method. Problem is, that you're doing quite a few other things in here. Callers of this method would expect a boolean value back, but not multiple side effects like printing to stdout and so on.

lib/rubycritic/cli/application.rb
```diff
@@ -14,12 +14,29 @@ def initialize(argv)
 
       def execute
         parsed_options = @options.parse
-        ::Rubycritic::CommandFactory.create(parsed_options).execute
+        result = ::Rubycritic::CommandFactory.create(parsed_options).execute
```
troessner: I believe "execute" should rather return a kind of Reporter initialized with the given options from the options_parser. And this reporter you can then ask: 

* "Is my score to low?" (with the reporter knowing if the flag has been set or not)
* "What should I return?"

You could check out how we do it in [`Reek`](https://github.com/troessner/reek/blob/master/lib/reek/cli/application.rb#L31) - as you can see, the "Application" still has access to the corresponding exit code even at the very end of the life time.

I'm thinking about something along the lines (Pseudo code obviously):

```Ruby
reporter = ::Rubycritic::CommandFactory.create(parsed_options).execute
reporter.exit_status
```

With the reporter knowing the options and then encapsulating this logic.

troessner: And since we're on it, let's drop the scope operator as well. Just

>> Rubycritic::CommandFactory.create...

is fine.

lib/rubycritic/cli/application.rb
troessner: Addendum: In reference to my comment above, this whole logic should not be here but in a Reporter.
Without any type check of course ;)

lib/rubycritic/cli/application.rb
troessner: Or in one sentence: This should actually be multiple extremely small methods.

troessner:
A general comment: Of course every pull request should have extensive test coverage - I assume you just didn't add any unit tests because you considered this as a "feeling out" process.
However, it would help to add 2 integration tests right from the start: 1 that tests our exit_status without "--minimum-score" and one that tests our exit_status with it. 
As much as I dislike cucumber, this would be very fitting here. We don't have it integrated at the moment, but I think this would be a good start. You can alway check out how we do it in [Reek](https://github.com/troessner/reek/tree/master/features/command_line_interface)

RobertoSchneiders:
Thanks for the help @troessner. 

I never worked with cucumber, so, it is the first time I configure and used it, but, I tried to do more or less the same of reek. I hope it is close to what it should be. 

I created the integration tests, I'm not sure if they are good enough, but it is a start. :)

About the reporter. I looked at the reek code and probably I don't completely understood your line of thinking. Rubicritic already have a reporter, but it is for json and html. Even so, I implemented the way I understood your idea, but, I had some problems. I decided to push the code anyway and start a discussion.

Problems with the actual implementation:
* I had to pass the full options hash to all commands
* I had to use inheritance to avoid a lot of duplication code in the commands (Don't know if worth it)
* Rubocop is complaining about a method that I'm using for inhertance override, which breaked the build (I'm looking for a solution)

I'm not english native, so, probably I made some typos, I will review this tomorrow.

Thanks for your patience.

features/command_line_interface/exit_status.feature
```diff
@@ -0,0 +1,20 @@
+Feature: Return the correct exit status codes
+  In order to use Rubycritic in Continuous Integration environments
+  Rubycritic returns the exit status code properly
+
+  Scenario: Status indicates a success when analyse a clean file
+    Given the clean file 'clean.rb'
+    When I run rubycritic clean.rb
+    Then the exit status indicates a success
```
troessner: Quick style thing: Let's use newlines between scenarios.

features/command_line_interface/exit_status.feature
```diff
@@ -0,0 +1,20 @@
...
+  Scenario: Status indicates a success when analyse a smelly file
+    Given the smelly file 'smelly.rb'
+    When I run rubycritic smelly.rb
+    Then the exit status indicates a success
+  Scenario: Status indicates a success when analyse a clean file with the minimum score option
+    Given the clean file 'clean.rb'
+    When I run rubycritic --minimum-score 100 clean.rb
+    Then the exit status indicates a success
+  Scenario: Status indicates an error when analyse a smelly file with the minimum score option
+    Given the smelly file 'smelly.rb'
```
troessner: Can we make this step more expressive by actually telling the caller what score this file has explicitly?
Thinking of something like

>> Given the smelly file 'smelly.rb' with a score of 115

features/command_line_interface/minimum_score.feature
```diff
@@ -0,0 +1,30 @@
+Feature: Brake if overall score is bellow minimum
+  In order to breake the Continuous Integration builds based on a score threshold
+  Rubycritic returns the exit status according with the score
+
+  Scenario: Status indicates an error when score below the minimum
+    Given the smelly file 'smelly.rb'
+    When I run rubycritic --minimum-score 100 smelly.rb
+    Then the exit status indicates an error
+  Scenario: Status indicates a success when score is above the minimum
+    Given the smelly file 'smelly.rb'
```
troessner: Same here - I, as the user have **no** idea what the real score is. So that we're using "50", "100" and so on is confusing to me. Let's put all of those numbers in the step definition so that it's obvious.

features/command_line_interface/exit_status.feature
```diff
@@ -0,0 +1,20 @@
...
+    When I run rubycritic --minimum-score 100 smelly.rb
+    Then the exit status indicates an error
```
troessner: I should have read the whole PR first ;) - this is an duplication to features/command_line_interface/minimum_score.feature (which you probably are aware of). I can see why you would do that but I still think it's redundant. I'd rather scrap this file completely and just use features/command_line_interface/minimum_score.feature. That you then can use this for CI is kind of implicitly.

features/command_line_interface/minimum_score.feature
```diff
@@ -0,0 +1,30 @@
...
+    """
+    Score: 100
+    """
+  Scenario: Prints a message informing the score is below the minimum
+    Given the empty file 'empty.rb'
+    When I run rubycritic --minimum-score 100 empty.rb
+    Then the output should contain:
+    """
+    Score (0.0) is below the minimum 100.0
+    """
```
troessner: Nice feature, that's exactly what I had in mind!

lib/rubycritic/cli/application.rb
```diff
@@ -9,17 +9,24 @@ class Application
       STATUS_ERROR   = 1
 
       def initialize(argv)
+        puts argv.class
```
troessner: I'm pretty you don't want this here ;)

lib/rubycritic/cli/options.rb
```diff
@@ -27,6 +27,10 @@ def parse
             @format = format
           end
 
+          opts.on("-s", "--minimum-score [MIN_SCORE]", "Set a minimum score") do |min_score|
+            @minimum_score = min_score.to_f
```
troessner: Don't we want to use an integer here?
And let's rather use proper constructor methods:

>>
[1] pry(main)> nil.to_i
=> 0
[2] pry(main)> Integer(nil)
TypeError: can't convert nil into Integer
from (pry):2:in `Integer'

lib/rubycritic/command_factory.rb
```diff
@@ -8,16 +8,16 @@ def self.create(options = {})
       case Config.mode
       when :version
         require "rubycritic/commands/version"
-        Command::Version.new
+        Command::Version.new(options_hash)
```
troessner: Let's not use [hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) - I think just calling it "options" is fine here.

lib/rubycritic/commands/status_reporter.rb
```diff
@@ -0,0 +1,45 @@
...
+      SUCCESS = 0
+      SCORE_BELOW_MINIMUM = 1
+
+      def initialize(options)
+        @options = options
+        @status = SUCCESS
+      end
+
+      def score=(score)
+        return if score.nil?
```
troessner: Please see my comment [here](https://github.com/whitesmith/rubycritic/pull/99/files#diff-c017e99a0be9d85fac1d8d58af3382a2R31)

lib/rubycritic/commands/status_reporter.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        @score = score.to_f
+        update_status
+        update_status_message
```
troessner: That's confusing since it's not obvious what the difference between those 2 methods.
Additionally, you're decoupling something that **must** be coupled, right? There's no point in updating the status without updating the message?
I'd rather merge "update_status_message" with "update_status"

troessner:
Overall: Very nice! This is going in the right direction!
Let's address all comments and then I'm confident we can get this one merged middle of next week ;)

lib/rubycritic/cli/application.rb
RobertoSchneiders: Ooops!! :grimacing:

features/command_line_interface/exit_status.feature
RobertoSchneiders: I opted for doing this because I was afraid that the score for the smelly code can change overtime, and this test will break frequently. That's not the case or this is not a problem?

lib/rubycritic/command_factory.rb
RobertoSchneiders: I haven't named that variable, but I agree, I will fix this. :+1:

lib/rubycritic/commands/status_reporter.rb
RobertoSchneiders: Nice catch.

lib/rubycritic/cli/application.rb
troessner: A wiser man than me once gave me a tip I still follow up until this day: Never indent your debugging statements. That way they'll always stick out like a sore thumb ;)

features/command_line_interface/exit_status.feature
troessner: >> That's not the case or this is not a problem?

Actually I think this is good because that you actually realise that something has changed.
Check out [this feature](https://github.com/troessner/reek/blob/master/features/samples.feature) in `Reek`. 
It is huge. We could  easily make that a lot thinner and remove some duplication. We don't. Because if we change something deep down and do NOT expect anything to change in `Reek`s behaviour and all of a sudden we have some changes there we know something went south.
Also, I'd take readability over churn any day, so let's go for readability ;)

RobertoSchneiders:
@troessner I made some changes. 

At some point I breaked the help functionality and I only noticed that by lucky in a manual test, so, I created a feature test for it. :)

RobertoSchneiders:
A test is breaking only on Travis, I'm trying to figure it out..

RobertoSchneiders:
It was breaking because of a change on the master branch. Fixed.

features/command_line_interface/minimum_score.feature
```diff
@@ -0,0 +1,39 @@
+Feature: Brake if overall score is bellow minimum
```
Onumis: typo: `Break`

features/command_line_interface/options.feature
```diff
@@ -0,0 +1,35 @@
...
+                                             html (default)
+                                             json
+                                             console
+          -s, --minimum-score [MIN_SCORE]  Set a minimum score
+          -m, --mode-ci                    Use CI mode (faster, but only analyses last commit)
+              --deduplicate-symlinks       De-duplicate symlinks based on their final target
+              --suppress-ratings           Suppress letter ratings
+          -v, --version                    Show gem's version
+          -h, --help                       Show this message
+
```
troessner: Very nice!

lib/rubycritic/command_factory.rb
```diff
@@ -3,21 +3,24 @@
...
-        Command::Help.new(options.help_text)
+        Command::Help
       when :ci
         require "rubycritic/commands/ci"
-        Command::Ci.new(options_hash[:paths])
+        Command::Ci
       else
         require "rubycritic/commands/default"
-        Command::Default.new(options_hash[:paths])
+        Command::Default
```
troessner: Nice refactoring!

lib/rubycritic/commands/status_reporter.rb
```diff
@@ -0,0 +1,43 @@
+module Rubycritic
+  module Command
+    class StatusReporter
+      attr_reader :status, :status_message
+      SUCCESS = 0
+      SCORE_BELOW_MINIMUM = 1
```
troessner: Lets use those constants in our features above instead of the magic numbers  0 and 1 directly.

troessner:
Great job! One last comment from my side and then lets get this one merged!

lib/rubycritic/commands/status_reporter.rb
RobertoSchneiders: :+1:

troessner:
Almost there! Now let's go the last step: Please squash your pull request into  one single commit and then I'll happily merge this one!

RobertoSchneiders:
@troessner Done :)

.travis.yml
```diff
@@ -13,4 +13,5 @@ matrix:
     - rvm: jruby
   fast_finish: true
 
+before_install: "gem update bundler"
```
troessner: Wait, why this change here? I fail to see how this belongs to this branch?

lib/rubycritic/commands/base.rb
```diff
@@ -0,0 +1,16 @@
...
+module Rubycritic
+  module Command
+    class Base
+      def initialize(options)
+        @options = options
+        @status_reporter = Rubycritic::Command::StatusReporter.new(@options)
+      end
+
+      def execute
+        @status_reporter
```
troessner: I had an uneasy feeling about this in the first review already and on second thought I don't think we should do this:

You're using the template pattern here where you overwrite `execute` in subclasses. The problem with the way do it right now is that you're creating an undocumented, obscure dependency here - corresponding methods in subclasses **must** call `super`, because otherwise the return type changes signifcantly. And not only that, they need to call `super` in a certain order as well.

Let's not do this, this will confuse ourselves in 4 weeks from now. How about just raising NotImplementedError here and then return "@status_report" in all `executes` below?

.travis.yml
troessner: Ah, I see where this is coming from. Pardon my french, but you fucked up the rebase.
This change and a **lot** of other changes in this PR are coming from the one [with the ConsoleReport](https://github.com/whitesmith/rubycritic/pull/91) we recently merged.
I'm afraid you'll have to remove all changes that slipped into your rebase before we can merge this.
The joys of git ;)

lib/rubycritic/core/smell.rb
```diff
@@ -8,7 +8,7 @@ class Smell
 
     attribute :context
     attribute :cost
-    attribute :locations
+    attribute :locations, Array, :default => []
```
troessner: Comes from https://github.com/whitesmith/rubycritic/pull/91

lib/rubycritic/generators/text/list.rb
```diff
@@ -0,0 +1,43 @@
+require "colorize"
```
troessner: Changes below come from [here](https://github.com/whitesmith/rubycritic/pull/91) as well.

troessner:
I'm afraid this is not ready to merge. 2 things:

1.) My comment in regards to lib/rubycritic/commands/base.rb - that should be easy to fix though.
2.) The second problem is more severe: You seemed to have mixed up your code with [the ConsoleReport PR](https://github.com/whitesmith/rubycritic/pull/91) - we need to remove all the code that came in from here.

RobertoSchneiders:
Sorry about the rebase @troessner , I forgot to remove that commit. 

I made the refactor about the execute method.

Onumis:
Looks good to me :+1:

.rubocop.yml
```diff
@@ -517,6 +518,7 @@ Style/TrivialAccessors:
     - to_str
     - to_s
     - to_sym
+    - execute
```
RobertoSchneiders: I removed this, but for some reason the diff is not updated..

troessner:
There seems to be an ongoing fight between you and git - looking at the current diff, we moved one step forward but two back ;)

1.) The `execute` method is back to calling `super` - looks like your latest changes are lost
2.) Like before: we need to squash this to one commit, right now its 18

RobertoSchneiders:
embarassing... I pushed the wrong branch. Sorry again! Fixed! 

Build failed, but, failed in `Updating installed gems`, seems to be something wrong with bundler. I can't restart the builds.

troessner:
Yes, seems like rubygems is the problem here. I restarted the build twice already :(
Let's retry tomorrow.

troessner:
Travis is green, merging this one. Great job, thanks!

RobertoSchneiders:
Thanks for your help (and for the patience) @troessner, sorry for the git mess. It was a great experience, learned a lot.

Issue #94, opened 2016-01-07
Analysers running twice in some cases


jbodah:
Haven't been able to dig into this yet, but it's been happening a lot lately. Will report more as it comes up. Figured I'd post in case others were experiencing this. Seems to happen whenever I destroy the tmp folder:

```
Bodah@jbodah-local 2.1.3p242 ~/repos/rubycritic (master) $ rm -rf tmp
Bodah@jbodah-local 2.1.3p242 ~/repos/rubycritic (master) $ bin/rubycritic lib
running flay smells

running flog smells
..................................................
running reek smells
..................................................
running complexity
..................................................
running attributes
..................................................
running churn
..................................................
running flay smells

running flog smells
..................................................
running reek smells
..................................................
running complexity
..................................................
running attributes
..................................................
running churn
..................................................
New critique at file:///Users/Bodah/repos/rubycritic/tmp/rubycritic/overview.html
Bodah@jbodah-local 2.1.3p242 ~/repos/rubycritic (master) $ vim lib/rubycritic/analysers
analysers/           analysers_runner.rb
Bodah@jbodah-local 2.1.3p242 ~/repos/rubycritic (master) $ vim lib/rubycritic/analysers_runner.rb
Bodah@jbodah-local 2.1.3p242 ~/repos/rubycritic (master) $ rm -rf tmp
Bodah@jbodah-local 2.1.3p242 ~/repos/rubycritic (master) $ bin/rubycritic lib

[22, 31] in /Users/Bodah/repos/rubycritic/lib/rubycritic/analysers_runner.rb
   22:       @paths = paths
   23:     end
   24:
   25:     def run
   26:       require 'rubygems'; require 'byebug'; byebug
=> 27:       ANALYSERS.each do |analyser_class|
   28:         analyser_instance = analyser_class.new(analysed_modules)
   29:         puts "running #{analyser_instance}"
   30:         analyser_instance.run
   31:       end
(byebug) require 'ap'
true
(byebug) ap caller
[
    [ 0] "/Users/Bodah/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/byebug-8.2.1/lib/byebug/helpers/eval.rb:59:in `eval'",
    [ 1] "/Users/Bodah/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/byebug-8.2.1/lib/byebug/helpers/eval.rb:59:in `safe_eval'",
[...]

PR #100, opened 2016-01-16
Open with browser

superiorlu:


superiorlu:
hi, i add a new feature: the report will be open with browser after generate report.

lib/rubycritic/browser.rb
```diff
@@ -0,0 +1,22 @@
+require "rubycritic/configuration"
+require "rubycritic/platforms/linux"
+require "rubycritic/platforms/darwin"
+
+module Rubycritic
+  class Browser
+    attr_reader :browser_name
```
troessner: Why not just "name"? The "browser_" seems unnecessary since it is scoped to browser already.

lib/rubycritic/browser.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    def platform
+      [Rubycritic::Platforms::Linux, Rubycritic::Platforms::Darwin].find(&:can_open?)
+    end
+
+    def open(report_path)
+      platform && platform.new(browser_name, report_path).open
+    end
+
```
troessner: No newline here pls.

lib/rubycritic/commands/default.rb
```diff
@@ -12,7 +13,8 @@ def initialize(paths)
       end
 
       def execute
-        report(critique)
+        return report(critique) if Config.format == :json
```
troessner: This will fail soon since a PR with a consolereport ist almost done.
Also I feel a 

```Ruby
case Config.format
when ....
end
```

is more expressive since it makes it clear that we are choosing paths based on format whereas your "return" makes it seem like json would be a special case.

lib/rubycritic/commands/default.rb
```diff
@@ -20,6 +22,15 @@ def critique
         RevisionComparator.new(@paths).set_statuses(analysed_modules)
       end
 
+      def open(report_path = nil)
```
troessner: ControlArgumentSmell. Lets move this whole logic to the caller since he already knows what to do.

lib/rubycritic/generators/html_report.rb
```diff
@@ -18,6 +18,7 @@ def generate_report
         create_directories_and_files
         copy_assets_to_report_directory
         puts "New critique at #{report_location}"
+        report_location
```
troessner: Why this change?

lib/rubycritic/platforms/base.rb
```diff
@@ -0,0 +1,38 @@
...
+      def self.can_open?
+        false
+      end
+
+      def browser
+        raise "not implement."
+      end
+
+      def open
+        raise "not implement."
```
troessner: You can replace this with a proper ruby exception:

```Ruby
raise NotImplementedError
```

lib/rubycritic/platforms/base.rb
```diff
@@ -0,0 +1,38 @@
...
+  module Platforms
+    class Base
+      attr_reader :browser_name, :report_path
+
+      def initialize(browser_name, report_path)
+        @browser_name = browser_name
+        @report_path  = report_path
+      end
+
+      def self.can_open?
```
troessner: Why is this a class method? Feels awkward...

troessner: And shouldnt we raise here as well?

lib/rubycritic/browser.rb
superiorlu: yes, i think this no need with 'browser_', i will modify it.

lib/rubycritic/platforms/darwin.rb
```diff
@@ -0,0 +1,27 @@
...
+
+module Rubycritic
+  module Platforms
+    class Darwin < Rubycritic::Platforms::Base
+      def self.can_open?
+        platform =~ /darwin/
+      end
+
+      def application
+        if browser_name =~ /chrome/i
```
troessner: We are mixing code and data here - how about just using a hash here and get rid of those ugly "if"s?

lib/rubycritic/platforms/linux.rb
```diff
@@ -0,0 +1,29 @@
+require "rubycritic/platforms/base"
+
+module Rubycritic
+  module Platforms
+    class Linux < Rubycritic::Platforms::Base
```
troessner: You can drop the RubyCritic::Platforms part here and above.

lib/rubycritic/generators/html_report.rb
superiorlu: because it want to get report path before report will be open in Command::Default class , so there is another way?

lib/rubycritic/commands/default.rb
superiorlu: good suggestion.

lib/rubycritic/platforms/linux.rb
superiorlu: okay, i will drop it.

lib/rubycritic/cli/options.rb
```diff
@@ -39,6 +39,10 @@ def parse
             @suppress_ratings = true
           end
 
+          opts.on("-O ", "--open", "Open report with browser") do |open_with|
```
troessner: Let's add all the possibilities here a user can enter, otherwise it's just happy guessing ;)

lib/rubycritic/commands/default.rb
```diff
@@ -12,7 +13,12 @@ def initialize(paths)
       end
 
       def execute
-        report(critique)
+        case Config.format
+        when :json
+          report(critique)
+        else
+          open(report(critique))
+        end
```
troessner: Yes, nice!

troessner: Since [this one](https://github.com/whitesmith/rubycritic/pull/91) is close to merge, let's update this to:

```Ruby
when :json
  report(critique)
when :html
  open(report(critique))
else
  raise "Unknown output format #{Config.format}"
end
```

lib/rubycritic/platforms/base.rb
```diff
@@ -0,0 +1,39 @@
...
+      def browser
+        raise NotImplementedError.new("The #{self.class} class must implement the #{__method__} method.")
+      end
+
+      def open(_)
+        raise NotImplementedError.new("The #{self.class} class must implement the #{__method__} method.")
+      end
+
+      def application
+        raise NotImplementedError.new("The #{self.class} class must implement the #{__method__} method.")
```
troessner: Those error messages are very repetitive - given that "normal" user never see this message I'd rather go for just

```Ruby
raise NotImplementedError
```
with no explicit error message to cut down repetition and improve readability.

lib/rubycritic/generators/html_report.rb
troessner: Ah, yeah, I missed that, nevermind ;)

lib/rubycritic/commands/default.rb
```diff
@@ -20,6 +26,15 @@ def critique
         RevisionComparator.new(@paths).set_statuses(analysed_modules)
       end
 
+      def open(report_path)
```
troessner: This method will only be called when we use a html_formatter, right?
This means there will **always** be a report_path, won't it?
Cause then we can drop the argument check and delegate this call directly to `browser`.

lib/rubycritic/commands/default.rb
troessner: I just merged the PR mentioned above, please rebase against master and then lets add handling :console_report here.

superiorlu: nice.

superiorlu:
hi troessner. 
i find a question： when i run ```rubycritic -p new_path ```， it seemed like run twice.
qq20160117-0 2x

troessner:
Yes, that's a [known problem](https://github.com/whitesmith/rubycritic/issues/94).
So it didn't originate in your PR ;)
Meaning: Just ignore that and let's fix this separately.

superiorlu:
okay. we will fix it.

superiorlu:
on 15th Jan , i published a topic about ```rubycritic``` on``` ruby china``` forum.
 [https://ruby-china.org/topics/28746](https://ruby-china.org/topics/28746). 
it hope that more pepole can use ```rubycritic``` analysis ruby code.

troessner:
Ok, this looks very good now, love it!

Now to the last steps, we need to 

[1] rebase against master since we have merge conflicts 
[2] squash to one commit
[3] final review

[1] and [2] are up to you, I'd tackle 3 then.
Please write a comment here when you're finished with [1] and [2] since github doesnt send out notifications for those events.

superiorlu:
Thanks for your help @troessner .
i will rebase and squash the commit.

superiorlu:
hi @troessner, i  have done it :)

lib/rubycritic/cli/options.rb
```diff
@@ -40,6 +40,16 @@ def parse
             @suppress_ratings = true
           end
 
+          opts.on("-O ", "--open", [:'chrome', :firefox, :safari, :chromium],
```
troessner: 2 points:

1.) The quotation marks for

>> :'chrome'

are unnecessary and seem to be a remnant.

2.) Can we make the list of supported browsers a constant so that we can reference that here and [here](https://github.com/whitesmith/rubycritic/pull/100/files#diff-0710b2dc2bc4fa0cf5b4aa9a98870c48R12)

lib/rubycritic/commands/default.rb
```diff
@@ -12,7 +13,15 @@ def initialize(paths)
...
       def execute
-        report(critique)
+        case Config.format
+        when :json
+          report(critique)
+        when :html
+          report(critique)
+          open
+        else
+          raise "Unknown output format #{Config.format}"
```
troessner: This will now crash on our newly merged "ConsoleReport", won't it?
Did you rebase against master?

lib/rubycritic/platforms/base.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      def browser_hash
+        raise NotImplementedError
+      end
+
+      def platform
+        @platform ||= begin
+          if RUBY_PLATFORM =~ /java/
+            `uname -s`.downcase
```
troessner: Wouldn't this crash for somebody running JRuby on windows?
How about doing something like [here](http://stackoverflow.com/questions/11784109/detecting-operating-systems-in-ruby)?

lib/rubycritic/platforms/darwin.rb
```diff
@@ -0,0 +1,29 @@
...
+module Rubycritic
+  module Platforms
+    class Darwin < Base
+      def can_open?
+        platform =~ /darwin/
+      end
+
+      def browser_hash
+        @browser_hash ||= begin
+          hash = { :chrome => "Google Chrome", :firefox => "firefox", :safari => "safari" }
```
troessner: Should be a constant, see my other comments above.

lib/rubycritic/platforms/darwin.rb
```diff
@@ -0,0 +1,29 @@
+require "rubycritic/platforms/base"
+
+module Rubycritic
+  module Platforms
+    class Darwin < Base
+      def can_open?
+        platform =~ /darwin/
+      end
+
+      def browser_hash
```
troessner: No hungarian notation please. How about "browser_map"? ("map" is some kind of generic interface while "hash" is a concrete type).

lib/rubycritic/platforms/darwin.rb
```diff
@@ -0,0 +1,29 @@
...
+        @browser_hash ||= begin
+          hash = { :chrome => "Google Chrome", :firefox => "firefox", :safari => "safari" }
+          hash.default = "safari"
+          hash
+        end
+      end
+
+      def application
+        browser_hash[browser_name.to_sym]
```
troessner: The "to_sym" seems like an unnecessary method call, isn't it?

lib/rubycritic/platforms/darwin.rb
```diff
@@ -0,0 +1,29 @@
...
+
+module Rubycritic
+  module Platforms
+    class Darwin < Base
+      def can_open?
+        platform =~ /darwin/
+      end
+
+      def browser_hash
+        @browser_hash ||= begin
```
troessner: If you use a constant like suggested above:

```Ruby
BROWSER_MAP = {}.tap do |hash|
  hash = { :chrome => "Google Chrome", :firefox => "firefox", :safari => "safari" }
  hash.default = "safari"
end

**Comment by troessner** [comment]
@superiorlu couple of more comments I'm afraid, but we're almost good to merge this one, let's stay the course!

File: lib/rubycritic/platforms/darwin.rb (hunk 20 above)
Comment (superiorlu) [inline lib/rubycritic/platforms/darwin.rb]: :+1:

**Comment by superiorlu** [comment]
@troessner  okay, i wil do that with your suggestions. Thanks for your patience.

**Comment by superiorlu** [comment]
@troessner i have refactor the code and rebase the commit.
Thanks for your help.

File: test/lib/rubycritic/browser_test.rb [hunk 21]
```diff
@@ -0,0 +1,67 @@
...
+require "rubycritic/platforms/darwin"
+require "rubycritic/generators/html_report"
+
+describe Rubycritic::Browser do
+  before do
+    @report_path = "tmp/rubycritic/overview.html"
+    @browser = Rubycritic::Browser.new @report_path
+  end
+
+  describe "open report on diffrent platform" do
```
Onumis: typo: `different`

lib/rubycritic/platforms/darwin.rb
```diff
@@ -0,0 +1,26 @@
+require "rubycritic/platforms/base"
+
+module Rubycritic
+  module Platforms
+    class Darwin < Base
+      BROWSER_MAP = {}.tap do |hash|
+        hash.merge! :chrome => "Google Chrome", :firefox => "firefox", :safari => "safari"
```
Onumis: just to be sure, is it really `Google Chrome` instead of `google-chrome`? (I don't have a mac so, I can't assess this myself)

test/lib/rubycritic/browser_test.rb
```diff
@@ -0,0 +1,67 @@
...
+        @linux.browser.must_equal "google-chrome"
+      end
+
+      it "open with chromium browser " do
+        Rubycritic::Config.stubs(:open_with).returns(:chromium)
+        Rubycritic::Browser.any_instance.stubs(:platform).returns(Rubycritic::Platforms::Linux.new)
+        @linux = @browser.platform
+        @linux.browser.must_equal "chromium-browser"
+      end
+    end
```
Onumis: I think we can apply some DRY here:

```
let(:report_path) { "tmp/rubycritic/overview.html" }
let(:browser) { Rubycritic::Browser.new report_path }

before do
  Rubycritic::Browser.any_instance.stubs(:platform).returns(target_platform)
end

describe "open report on different platform" do
  context "on linux" do
    let(:target_platform) { Rubycritic::Platforms::Linux.new }
    it "open with default browser" do
      Rubycritic::Config.set
      platform.browser.must_equal "firefox"
    end
    #...  
  end
  context "on darwin" do
    let(:target_platform) { Rubycritic::Platforms::Linux.new }
    #...
  end
end
```

I haven't tested this, but it should work

features/command_line_interface/options.feature
```diff
@@ -29,6 +29,7 @@ Feature: Rubycritic can be controlled using command-line options
           -m, --mode-ci                    Use CI mode (faster, but only analyses last commit)
               --deduplicate-symlinks       De-duplicate symlinks based on their final target
               --suppress-ratings           Suppress letter ratings
+          -O, --open                       Open html report with browser: [:chrome, :chromium, :firefox, :safari]
```
Onumis: Maybe we should say here `(Leave blank to open system default)`?

lib/rubycritic/platforms/darwin.rb
superiorlu: @Onumis  yes, i test it in my mac osx 10.10.5.
 [http://superuser.com/questions/350309/how-can-i-open-a-url-in-google-chrome-from-the-terminal-in-os-x](http://superuser.com/questions/350309/how-can-i-open-a-url-in-google-chrome-from-the-terminal-in-os-x)

test/lib/rubycritic/browser_test.rb
superiorlu: nice. that looks better :)

features/command_line_interface/options.feature
superiorlu: on linux, firefox is default. 
on mac osx, safari is default.
so i think default may be not important. just run ```rubycritic``` :smile:

test/lib/rubycritic/browser_test.rb
Onumis: Sure, push those changes and fix the typo on "diffrent" and it'll be good to go :)

superiorlu: Okay, thanks. @Onumis

superiorlu:
Hi @Onumis i have fixed it and rebase commit. Thanks.

lib/rubycritic/platforms/base.rb
```diff
@@ -0,0 +1,47 @@
...
+          case host_os
+          when /mswin|msys|mingw|cygwin|bccwin|wince|emc/
+            :windows
+          when /darwin|mac os/
+            :darwin
+          when /linux/
+            :linux
+          when /solaris|bsd/
+            :unix
+          end
```
troessner: Yes, that is a LOT nicer now :)

lib/rubycritic/platforms/base.rb
```diff
@@ -0,0 +1,47 @@
...
+        end
+      end
+
```
troessner: This will break the build after rebase. Now newline here pls.

lib/rubycritic/platforms/darwin.rb
```diff
@@ -0,0 +1,26 @@
...
+      end
+
+      def can_open?
+        platform == :darwin
+      end
+
+      def application
+        BROWSER_MAP[browser_name]
+      end
+      alias_method :browser, :application
```
troessner: This will also break the build after rebase because rubocop will complain about that, sorry ;)
Should be `alias`.

lib/rubycritic/platforms/darwin.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      def open(report_path)
+        system("open -a '#{application}' #{report_path}")
+      end
+    end
+
```
troessner: See above.

troessner:
Actually you can scratch my comments about breaking the build since my PR is not merged yet :)

troessner:
Merged! Great job, thanks!

superiorlu:
Thanks @troessner @Onumis i have learned more than i expected. Thanks your help and patience.
This is my first project which contributed  to in github,  your advice makes me more confident. Thanks again. :)

PR #105, opened 2016-01-21
Bump version to 2.6.0.

troessner:


nijikon:
Can we also bump reek to 3.9.0?

troessner:
Good idea! I need to bump rubocop first though since this is relying on an old version of parser. 
Looking into this now.

nijikon:
@troessner Sweet, thanks!

PR #106, opened 2016-01-22
[WIP] Slim down RuboCop configuration and bump RuboCop and Reek.

troessner:
Well, well, well, look at all those changes! I always wondered how to insert dog pictures in github pull requests, now would actually be the time! 

doge

Ok, where was I? Ah, right, this pull request fixes several issues:

1.) Before this PR our rubocop.yml was **huge**. This is just plain wrong. The point of a styleguide is to have as minimal deviations as possible. Otherwise you can just drop the whole damn thing since it's more confusing than helpful. I kind of slimmed it down to the bare minimum. 
2.) I fixed 98% of all problems (or in most cases: let Rubocop fix it) that arose from [1] and created a small rubocop_todo.yml we need to fix in the future
3.) I bumped RuboCop to 0.36 which is the latest version that is running the latest parser which will give us Ruby 2.3 compatibility, at least for the tools that depend on parser (like Reek)
4.) I bumped Reek to 3.9 which is the latest version that is running the latest parser

P.S.: I know this PR looks huge but most changes are trivial like ' vs " so let's get this bad boy reviewed!

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,7 @@ AllCops:
     - '**/Rakefile'
   Exclude:
     - 'test/samples/**/*'
+    - 'tmp/**/*'
   # By default, the rails cops are not run. Override in project or home
   # directory .rubocop.yml files, or by giving the -R/--rails option.
   RunRailsCops: false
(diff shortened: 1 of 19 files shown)


troessner:
Hold on with the review and let me fix those conflicts first ;)

troessner:
>> Pull request successfully merged and closed

Ok, something is not right with github. Neither did I merge this one nor is it merged. WTF.
Reverting this one. Sorry about that, I have no idea what happened here.

troessner:
Ok, somehow github's display of this one is screwed up. Never managed to do that :laughing: 
I'll delete this one and create a new one from scratch in the hope that this doesn't show the same problems.

Onumis:
Ah! I blame the doge :P

PR #111, opened 2016-01-25
Changes the output score format to 2 decimals

RobertoSchneiders:
I added a round to the score on the status_reporter to be a litle more presentable on the output. It was printing 
```
Score: 97.95258620689656
```
now it will print only 2 decimals
```
Score: 97.95
```
WDYT?


Onumis:
Makes sense! Thanks for paying attention to the details!

PR #119, opened 2016-02-08
reek 3.10.1 and rubocop 0.37.0

nijikon:


lib/rubycritic/commands/base.rb
```diff
@@ -9,7 +9,7 @@ def initialize(options)
       end
 
       def execute
-        fail NotImplementedError
+        raise NotImplementedError
```
troessner: Not the biggest fan of this since I really like the semantic difference between "fail" and "raise". 
However, sticking a reasonable style guide project wide is more important than my personal taste here.
@Onumis @lucasmazza wdyt?
If you guys don't care we can just go with "raise" for all-the-things.

lib/rubycritic/revision_comparator.rb
```diff
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
```
troessner: Newline please here and below.

troessner:
Couple of small comments and then lgtm ;)

nijikon:
@troessner fixed what you pointed out, waiting for your decision on raise vs fail :)

lib/rubycritic/commands/base.rb
Onumis: I do appreciate using `fail` to create an exception and `raise` to re-raise it and I believe Rubocup also suggests that.
But I often overlook these things and just type `raise` everywhere.

That being said, and since we're trying to help people write better ruby code, we should follow the community trends.
I vote to stick with `fail` here.

Onumis: fwiw: https://github.com/bbatsov/ruby-style-guide#prefer-raise-over-fail

Onumis: After reading through that last link, it turns out the tides are shifting!
These are quite recent changes but it suggests to move away from fail - Rubocop 0.37.0 does just that.

So I changed my mind here, we should stick with `raise`.

Personal preferences are over-rated :P

troessner: >> These are quite recent changes but it suggests to move away from fail - Rubocop 0.37.0 does just that.

Yep, exactly :laughing: 

>> Personal preferences are over-rated

I assume @lucasmazza agrees with that, so let's get this merged ;)

troessner:
@nijikon I'd like to merge this one, there's just one thing that's keeping me from doing so: Can you please squash your commits to 1 whose commit message is the title of this pull request?

nijikon:
@troessner Done :)

troessner:
Merged, thanks!

PR #126, opened 2016-02-20
Smell documentation link

danielmbarlow:
Added links on smells to Reek documentation on Github in the code screen. All comments are welcome. We don't use Minitest at work, so let me know if there's anything you want me to change with the tests.

### DIFF
--- a/lib/rubycritic/core/smell.rb
+++ b/lib/rubycritic/core/smell.rb
@@ -46,6 +46,10 @@ module Rubycritic
       to_h.to_json(*a)
     end
 
+    def doc_url
+      "https://github.com/troessner/reek/blob/master/docs/#{dasherized_type}.md"
+    end
+
     def hash
       state.hash
     end
@@ -54,6 +58,12 @@ module Rubycritic
 
     def state
       [@context, @message, @score, @type]
+    end
+
+    private
+
+    def dasherized_type
+      type.gsub(/(?<!^)([A-Z])/, '-\1')
     end
   end
 end
--- a/test/lib/rubycritic/core/smell_test.rb
+++ b/test/lib/rubycritic/core/smell_test.rb
@@ -69,4 +69,18 @@ describe Rubycritic::Smell do
       smell1.must_equal smell2
     end
   end
+
+  describe '#doc_url' do
+    it 'handles one word type names' do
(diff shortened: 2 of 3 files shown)


lib/rubycritic/core/smell.rb
```diff
@@ -46,6 +46,14 @@ def to_json(*a)
       to_h.to_json(*a)
     end
 
+    def dasherized_type
```
troessner: Lets make this method private, it doesnt have to be public.

test/lib/rubycritic/core/smell_test.rb
```diff
@@ -69,4 +69,27 @@
       smell1.must_equal smell2
     end
   end
+
+  describe '#dasherized_type' do
```
troessner: I believe we should test this method implicitly through doc_url  - how about moving those tests there?

troessner:
Btw. this functionality (obviously) exists in Reek as well, unfortunately it's not really well exposed:

```Ruby
warning = FactoryGirl.build(:smell_warning)
Reek::Report::WikiLinkWarningFormatter.new.send :explanatory_link, warning

>>
=> "https://github.com/troessner/reek/blob/master/docs/Feature-Envy.md"

It probably would make sense to expose this in a better way. But that would require a little more sophisticated approach than just making the method public ;)

I guess for now it's better to go with your solution.
Two small comments and the lgtm

danielmbarlow:
Hi Timo, I made the changes you suggested, squashed my commits (as I think your CONTRIBUTE.md suggests), and rebased on master. Let me know if you need anything else done for this PR.

troessner:
Looks good now!  But travis rightfully complains about rubocop violations ;)

danielmbarlow:
Hi Timo, how are we looking now?

troessner:
Looks good now! travis is still red because of jruby but that shouldn't be an issue after https://github.com/whitesmith/rubycritic/pull/128 -> merged!

troessner:
Btw. extremely awesome pull request, a small change but one that increases convenience significantly, thanks for that!

danielmbarlow:
You are welcome. I look forwards to using it. Thanks for Rubycritic!

Onumis:
This is so awesome! Thanks @danielmbarlow

PR #127, opened 2016-02-20
Make RubyCritic usable as Rake task.

### LINKED ISSUE #95: Make RubyCritic usable as Rake task
Coming from #85 I'd like to take this over and start fresh.

troessner:
Preliminary PR - please don't comment yet.
Fixes #95

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,45 @@ If you're fond of Guard you might like [guard-rubycritic][4]. It automatically a
 
 For continuous integration, you can give [Jenkins CI][5] a spin. With it, you can [easily build your own (poor-man's) Code Climate][6]!
 
+Rake Task
+---------
+
+You can use RubyCritic as Rake command in its most simple form like this:
+
+```ruby
+require "rubycritic/rake_task"
+
+Rubycritic::RakeTask.new
+```
+
+A more sophisticated Rake task that would make use of all available configuration options could look like this:
+
+```ruby
+RubyCritic::RakeTask.new do |task|
+  # Name of RubyCritic task. Defaults to :rubycritic.
+  task.name    = 'something_special'
+
+  # Glob pattern to match source files. Defaults to FileList['.'].
+  task.paths   = FileList['vendor/**/*.rb']
+
+  # You can pass all the options here in that are shown by "rubycritic -h" except for
+  # "-p / --path" since that is set separately. Defaults to ''.
+  task.options = '--mode-ci --format json'
+
+  # Defaults to false
+  task.verbose = true
+end
+```
+
+RubyCritic will try to open the generated report with a browser by default. If you don't want this
+you can prevent this behaviour by setting the options correspondingly:
+
+```ruby
+RubyCritic::RakeTask.new do |task|
+  task.options = '--no-browser'
+end
+```
+
 Compatibility
 ---------------
 
--- a/lib/rubycritic/rake_task.rb
+++ b/lib/rubycritic/rake_task.rb
@@ -1,0 +1,70 @@
+require 'rake'
+require 'rake/tasklib'
+require 'English'
+require 'rubycritic/cli/application'
+
+module Rubycritic
+  #
+  # A rake task that runs RubyCritic on a set of source files.
+  #
+  # This will create a task that can be run with:
+  #
+  #   rake rubycritic
+  #
+  # Example:
+  #
+  #   require 'rubycritic/rake_task'
+  #
+  #   RubyCritic::RakeTask.new do |task|
+  #     task.paths = FileList['lib/**/*.rb', 'spec/**/*.rb']
+  #   end
+  #
+  # :reek:TooManyStatements: { max_statements: 6 }
+  class RakeTask < ::Rake::TaskLib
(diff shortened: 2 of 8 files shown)


troessner:
Ready for review! cc @Onumis @lucasmazza

troessner:
Jruby fails the build. Like usual. I thought we had made Jruby an "allowed failure", but apparently not.
I'll whip up another pull request to address this so this one gets green after rebasing.

lib/rubycritic/rake_task.rb
```diff
@@ -0,0 +1,75 @@
...
+    attr_reader :name, :paths, :verbose, :options
+
+    def define_task
+      desc 'Run RubyCritic'
+      task(name) { run_task }
+    end
+
+    def run_task
+      puts "\n\n!!! Running #{name} rake command: #{command}\n\n" if verbose
+      system(*command)
```
lucasmazza: Can't we dog food our own Ruby API instead of going the `system` route? We could start with something like the following (we would still need to support the `path`, `minimum-score` and other options):

```ruby
command = Rubycritic::Command::Default.new(paths: paths)
reporter = command.execute
print(reporter.status_message)
exit reporter.status
```

Or go straight down the `Rubycritic::Cli::Application.new(*(options_as_arguments + paths))` way.

README.md
```diff
@@ -116,6 +116,38 @@ If you're fond of Guard you might like [guard-rubycritic][4]. It automatically a
...
+
+```ruby
+require "rubycritic/rake_task"
+
+Rubycritic::RakeTask.new do |task|
+  task.verbose = true
+end
+```
+
+An more sophisticated Rake task that would make use of all available configuration options could look like this:
```
Onumis: typo: "A more"

lib/rubycritic/rake_task.rb
Onumis: I agree, that seems like the ideal scenario.
Even though we don't have a well defined API (we're working on it I promise!) we should use what we have.

troessner: Agreed ;)
I'll update the PR to use

>> Or go straight down the Rubycritic::Cli::Application.new(*(options_as_arguments + paths)) way.

I'll ping you as soon as it's ready for re-review.

troessner:
@Onumis && @lucasmazza addressed all comments, ready for re-review!

rubycritic.gemspec
```diff
@@ -35,4 +35,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'minitest', '~> 5.3'
   spec.add_development_dependency 'mocha', '~> 1.1'
   spec.add_development_dependency 'rubocop', '0.37.2'
+  spec.add_development_dependency 'pry-byebug'
```
troessner: Not really necessary for this PR but I believe we should have this in our dev deps anyway and I am a big believer of the boyscout rule ;)

lib/rubycritic/rake_task.rb
```diff
@@ -0,0 +1,69 @@
...
+    # the rubycritic command to stdout. Defaults to false.
+    attr_writer :verbose
+
+    # You can pass all the options here in that are shown by "rubycritic -h" except for
+    # "-p / --path" since that is set separately. Defaults to ''.
+    attr_writer :options
+
+    def initialize(name = :rubycritic)
+      @name    = name
+      @paths   = FileList['lib/**/*.rb']
```
lucasmazza: Our default is `'.'` if no path is given to the CLI - makes sense to preserve that here instead of this new value?

lucasmazza:
Left a minor comment on the `paths` argument but everything else is gold. It's hard to admit that I'm starting the like to use of Cucumber/Aruba here.

lib/rubycritic/rake_task.rb
troessner: Good catch, fixed!

troessner:
>> Left a minor comment on the paths argument but everything else is gold.

Updated! Merge?

>>  It's hard to admit that I'm starting the like to use of Cucumber/Aruba here.

I feel the same :laughing:

Onumis:
Lgtm!

lucasmazza:
We gotta rebase this one to handle conflicts with #130 changes.

lib/rubycritic/rake_task.rb
```diff
@@ -0,0 +1,70 @@
...
+  #
+  # Example:
+  #
+  #   require 'rubycritic/rake_task'
+  #
+  #   RubyCritic::RakeTask.new do |task|
+  #     task.paths = FileList['lib/**/*.rb', 'spec/**/*.rb']
+  #   end
+  #
+  # :reek:TooManyStatements: { max_statements: 6 }
```
troessner: Need to disable this warning via source code comment since "exclude" in config files doesn't work for detectors on class base. We just recently discovered [this bug](https://github.com/troessner/reek/issues/816) in Reek, I'll work on a fix soon.

troessner:
Rebased and ready to merge! @Onumis / @lucasmazza I never merge my own pull requests, could one of you please go ahead if you're ok with the PR?

lucasmazza:
:v:

PR #128, opened 2016-02-21
Make JRuby an allowed failure.

troessner:
I believe we missed that when we recently worked on our travis config. Jruby should always be an allowed failure. This will also make #127 green.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ env:
 matrix:
   allow_failures:
     - rvm: ruby-head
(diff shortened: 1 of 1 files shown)


troessner:
@Onumis @lucasmazza objections to merge?

lucasmazza:
@troessner no :shipit:

troessner:
>> @troessner no :shipit:

Ok, then I'll merge this one myself even if that makes my toe nails curl up ;)

Onumis:
:+1: Thank you!

PR #130, opened 2016-02-28
Introduce Reek to our CI build.

troessner:
Kind of ironic that a gem that heavily relies on Reek to tell other people how their code scores does not use it itself. 
On the other Reek itself didn't use Reek for its own CI build until a year ago or so. I see a pattern :laughing:

### DIFF
--- a/.todo.reek
+++ b/.todo.reek
@@ -1,0 +1,133 @@
+---
+IrresponsibleModule:
+  exclude:
+  - Rubycritic::Analyser::Attributes
+  - Rubycritic::Analyser::Churn
+  - Rubycritic::Analyser::Complexity
+  - Parser::AST::Node
+  - Rubycritic::AST::EmptyNode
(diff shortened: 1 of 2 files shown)


troessner:
Addendum: I also included an initial .todo.reek file generated by one of our latest addition to Reek to, a "todo" [flag](https://github.com/troessner/reek/pull/852) similar to Rubocops --autogen-config 
This makes our Reek adoption easier and we can work on this file gradually whenever we feel like it.

Onumis:
I like the idea, but ideally we should go all the way and introduce RubyCritic to the CI.
We now have a cmd line output so we should be fine.
We can still maintain that .todo.reek so I'm gonna merge this now.

troessner:
>> I like the idea, but ideally we should go all the way and introduce RubyCritic to the CI.

Agreed. We just need to figure out what thresholds to apply ;)

Issue #117, opened 2016-02-05
Rubycritic throws errors with Ruby 2.3 syntax


seanhandley:
This is due to this commit in the `ruby_parser` library not yet being released: https://github.com/seattlerb/ruby_parser/commit/da22d1e256305dff2bb549e9f13fcf9f69c43bb2

Flay and Flog both depend on an earlier version of `ruby_parser` that doesn't have support.

```ruby
# foo.rb
Object&.new
```

```
$ rubycritic
  foo.rb:1 :: parse error on value "." (tDOT)
  skipping foo.rb
running flay smells

running flog smells
ERROR: parsing ruby file foo.rb
Racc::ParseError: foo.rb:1 :: parse error on value "." (tDOT) at:
  /Users/sean/.rbenv/versions/2.3.0/lib/ruby/2.3.0/racc/parser.rb:528:in `on_error'
  /Users/sean/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/ruby_parser-3.7.3/lib/ruby_parser_extras.rb:1125:in `on_error'
  /Users/sean/.rbenv/versions/2.3.0/lib/ruby/2.3.0/racc/parser.rb:259:in `_racc_do_parse_c'
  /Users/sean/.rbenv/versions/2.3.0/lib/ruby/2.3.0/racc/parser.rb:259:in `do_parse'
  /Users/sean/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/ruby_parser-3.7.3/lib/ruby_parser_extras.rb:1037:in `block in process'
.
running reek smells
.
running complexity
ERROR: parsing ruby file foo.rb
Racc::ParseError: foo.rb:1 :: parse error on value "." (tDOT) at:
  /Users/sean/.rbenv/versions/2.3.0/lib/ruby/2.3.0/racc/parser.rb:528:in `on_error'
  /Users/sean/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/ruby_parser-3.7.3/lib/ruby_parser_extras.rb:1125:in `on_error'
  /Users/sean/.rbenv/versions/2.3.0/lib/ruby/2.3.0/racc/parser.rb:259:in `_racc_do_parse_c'
  /Users/sean/.rbenv/versions/2.3.0/lib/ruby/2.3.0/racc/parser.rb:259:in `do_parse'
[...]

PR #131, opened 2016-02-28
Update analyzer configuration in README.

### LINKED ISSUE #77: Document how RubyCritic configures its analyzers
Issue originated in #75

troessner:
Fixes #77

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,10 +104,11 @@ $ rubycritic --help
 Analyzer Configuration
 -------------------------
 
-* `Reek`: `RubyCritic` just uses `Reek`'s default [configuration loading mechanism](https://github.com/troessner/reek#configuration-file).
-
-This section is work in progress at the moment and is missing the other analyzers that RubyCritic supports.
-You can expect this to be updated soon.
+* [`Reek`](https://github.com/troessner/reek): `RubyCritic` utilizes `Reek`'s default [configuration loading mechanism](https://github.com/troessner/reek#configuration-file).
+  This means that if you have an existing `Reek` configuration file, you can just put this into your
+  project root and `RubyCritic` will respect this configuration.
+* [`flay`](https://github.com/seattlerb/flay): We use `flay`'s default configuration.
+* [`flog`](https://github.com/seattlerb/flog): We use `flog`'s default configuration with a couple of [smaller tweaks](https://github.com/whitesmith/rubycritic/blob/master/lib/rubycritic/analysers/helpers/flog.rb#L5).
 
 Alternative Usage Methods
 -------------------------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -124,10 +125,7 @@ RubyCritic is supporting:
 * 2.0
 * 2.1
 * 2.2
-
-Note that we do not support Ruby 1.9 anymore.
```
troessner: With Ruby's 2.0 having reached the end of its lifetime I believe we seriously need to encourage people to use an up to date Ruby. I think we shouldn't litter our code and doc base with 1.9 references anymore.

README.md
```diff
@@ -124,10 +125,7 @@ RubyCritic is supporting:
...
-
-If you're still on ruby 1.9 you'll have to use the last version of RubyCritic that supports it, which would be [1.4.0](https://github.com/whitesmith/rubycritic/tree/v1.4.0).
+* 2.3
```
troessner: See my latest comment in #117

README.md
```diff
@@ -104,10 +104,11 @@ $ rubycritic --help
...
 
-* `Reek`: `RubyCritic` just uses `Reek`'s default [configuration loading mechanism](https://github.com/troessner/reek#configuration-file).
-
-This section is work in progress at the moment and is missing the other analyzers that RubyCritic supports.
-You can expect this to be updated soon.
+* [`Reek`](https://github.com/troessner/reek): `RubyCritic` utilizes `Reek`'s default [configuration loading mechanism](https://github.com/troessner/reek#configuration-file).
+  This means that if you have an existing `Reek` configuration file, you can just put this into your
+  project root and `RubyCritic` will respect this configuration.
+* [`flay`](https://github.com/seattlerb/flay): We use `flay`'s default configuration.
+* [`flog`](https://github.com/seattlerb/flog): We use `flog`'s default configuration with a couple of [smaller tweaks](https://github.com/whitesmith/rubycritic/blob/master/lib/rubycritic/analysers/helpers/flog.rb#L5).
```
troessner: And no, i have zero idea what those smaller tweaks mean since `flog` has not really documented those :laughing: 
Skimming through flogs codebase it looks "kind-of" default-ish though, so I'd rather be a little vague here than document flog's internal. @guilhermesimoes maybe you still remember why you added those tweaks i link to above?

troessner: BTW this flog configuration behaviour is nowhere spec'ed.

guilhermesimoes: > BTW this flog configuration behaviour is nowhere spec'ed.

It's kinda spec'ed on the [CLI code](https://github.com/seattlerb/flog/blob/258a1c5f0e3b356869804e47bd7c7c400a1488b6/lib/flog_cli.rb#L76).

Anyway, here are explanations for the options:

* `all` - Forces Flog to report scores on all classes and methods. Without this option Flog will only give results up to a certain threshold.

* `continue` - Makes it so that Flog does not abort when a ruby file cannot be parsed. I wish RubyCritic had this option so it could be properly delegated to the other gems.

* `methods` - Configures Flog to skip code outside of methods. It prevents Flog from reporting on the "methods" `private` and `protected`. It also prevents Flog from reporting on Rails methods like `before_action` and `has_many`.

README.md
guilhermesimoes: :+1:

PR #133, opened 2016-02-29
Specify runtime dependency on ruby_parser for ruby 2.3 support and other fixes

Onumis:
Related to https://github.com/whitesmith/rubycritic/issues/117#issuecomment-189880517

### DIFF
--- a/rubycritic.gemspec
+++ b/rubycritic.gemspec
@@ -25,6 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'flog', '4.3.2'
   spec.add_runtime_dependency 'reek', '3.11'
   spec.add_runtime_dependency 'parser', '~> 2.3'
+  spec.add_runtime_dependency 'ruby_parser', '~> 3.8'
   spec.add_runtime_dependency 'colorize'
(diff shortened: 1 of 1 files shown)


rubycritic.gemspec
```diff
@@ -25,6 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'flog', '4.3.2'
   spec.add_runtime_dependency 'reek', '3.11'
   spec.add_runtime_dependency 'parser', '~> 2.3'
+  spec.add_runtime_dependency 'ruby_parser', '>= 3.8'
```
troessner: But with this set up we would also pull in v4 which might be api breaking.
How about:

```Ruby
spec.add_runtime_dependency 'ruby_parser', [ '>= 3.8', ~> '3.8' ]
```

?

Onumis: Well, `ruby_parser` isn't following semantic versioning (I think) and I'd imagine this is the kind of gem that wouldn't ever introduce incompatible changes. But it can't hurt to be a bit strict on this...
Will change in a bit

Onumis:
Updated!

troessner:
Merged ;)

PR #134, opened 2016-03-05
Explain `flog` default configuration.

troessner:
Follow-up from #131 - adding @guilhermesimoes awesome explanation :)


Onumis:
:+1: great

PR #139, opened 2016-04-11
rubocop 0.39.0, reek 4.0.1, drop support for ruby 2.0

nijikon:


nijikon:
@troessner I assume that we want to support ruby 2.0?

Onumis:
Official support for Ruby 2.0 ended 2 months ago, and I believe Reek v4 drops support for it as well.
With this update we should also.

troessner:
PR and commit say "rubocop" but it upgrades Reek as well...we should fix that.

>> Official support for Ruby 2.0 ended 2 months ago, and I believe Reek v4 drops support for it as well.
With this update we should also.

Yep, good idea!

nijikon:
@troessner changed PR subject. There are 3 commits.

troessner:
Let's update our [compatibility](https://github.com/whitesmith/rubycritic#compatibility) section as well ;)

nijikon:
@troessner Done.

troessner:
Merged, thanks!

PR #141, opened 2016-04-12
Add simple .xml output generation.

emjei:
We need a way to generate simple .xml output for integration with CI. Also I noticed that output in JSON format does not generate a file, it only prints JSON output to a console. So I made that XML output generation would respect the -p flag (I should have probably added this to JSON as well).

### DIFF
--- a/lib/rubycritic/generators/xml_report.rb
+++ b/lib/rubycritic/generators/xml_report.rb
@@ -1,0 +1,24 @@
+require 'rubycritic/generators/xml/simple'
+
+module Rubycritic
+  module Generator
+    class XmlReport
+      def initialize(analysed_modules)
+        @analysed_modules = analysed_modules
+      end
+
+      def generate_report
+        report = generator.render
+
+        File.write "#{ Config.root }/report.xml", report
+        print report
+      end
+
+      private
+
+      def generator
+        Xml::Simple.new(@analysed_modules)
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


troessner:
Thanks for the PR, but this needs to be properly spec'ed :)

lib/rubycritic/generators/xml_report.rb
```diff
@@ -0,0 +1,24 @@
...
+        report = generator.render
+
+        File.write "#{ Config.root }/report.xml", report
+        print report
+      end
+
+      private
+
+      def generator
+        Xml::Simple.new(@analysed_modules)
```
troessner: I question this design even though it might be idiomatic for our current codebase. You are adding a hardcoded indirection from "base" to the "report" that offers no advantages but has the disadvantage of being less readable and confusing. If xml reporting was part of our domain i could see how we might have more than 1 strategy to generate xml reports but in its current state this reeks a bit of premature optimization.

I'd rather dissolve "base" into "xml_report" and keep it simple.

emjei: Hi.Sorry for the late response. Long story short, we needed even more changes like str

emjei: Hi. Sorry for the late response. Long story short, we needed even more changes like picking which analysers to run. This diverges much from the original code. So I think, it would be a better idea to just keep those changes inside our fork. Sorry for wasting your time.

PR #142, opened 2016-04-19
Generates report inside rails application

felipegruoso:
Changes the structure of HTML files to the same as a rails application.
Updates README file.


README.md
```diff
@@ -92,15 +92,40 @@ For a full list of the command-line options run:
...
+| `-p/--path`              | Sets the output directory (tmp/rubycritic by default)                                                     |
+| `-f'/--format [FORMAT]`  | Report smells in the given format: html (default), json, console                                          |
+| `-s/--minimum-score`     | Set a minimum score                                                                                       |
+| `--mode-ci`              | Uses CI mode (faster, but only analyses last commit)                                                      |
+| `--deduplicate-symlinks` | De-duplicate symlinks based on their final target                                                         |
+| `--suppress-ratings`     | Suppress letter ratings                                                                                   |
+| `--no-browser`           | Do not open html report with browser                                                                      |
+| `-r/-for-rails`          | Generates report inside your rails application. Ignores some given parameters: (-p, --path, -f, --format) |
```
troessner: Should be "--for-rails", shouldn't it?

README.md
```diff
@@ -92,15 +92,40 @@ For a full list of the command-line options run:
...
+
+If the -r option is choosen, it is possible build routes and dynamic controller to access it inside your rails app:
```
troessner: "possible build routes" -> "possible to build routes"

and 

>> and dynamic controller to access it inside your rails app

is confusing - what is "it"?

lib/rubycritic/command_factory.rb
```diff
@@ -4,9 +4,18 @@ module Rubycritic
   class CommandFactory
     def self.create(options = {})
       Config.set(options)
+      validate_rails_directory if Config.for_rails
```
troessner: Yeah, that is the thing that we need to watch for: All of a sudden our generic CommandFactory contains Rails specific code which decreases cohesion and increases coupling. How about creating commands/rails.rb?

lib/rubycritic/command_factory.rb
```diff
@@ -4,9 +4,18 @@ module Rubycritic
...
       command_class(Config.mode).new(options)
     end
 
+    def self.validate_rails_directory
+      raise ArgumentError, 'The option -r should be executed inside a rails app root folder' unless is_a_rails_app?
```
troessner: Then this check and the check below should be moved there as well.

lib/rubycritic/command_factory.rb
```diff
@@ -4,9 +4,18 @@ module Rubycritic
...
+    end
+
+    def self.is_a_rails_app?
+      File.exists?('Gemfile')
```
troessner: I'd say 90% of all modern non-rails-but-ruby have a Gemfile in their root directory so this check is actually a bug. I'd rather check for things that are only rails specific OR not care at all since checking this might always lead to edge case situations which will trip users up.

lib/rubycritic/configuration.rb
```diff
@@ -2,7 +2,8 @@ module Rubycritic
   class Configuration
     attr_reader :root
     attr_accessor :source_control_system, :mode, :format, :deduplicate_symlinks,
-                  :suppress_ratings, :open_with, :no_browser
+                  :suppress_ratings, :open_with, :no_browser, :assets_folder,
```
troessner: Same here with cohesion and coupling - all configuration values here are completely generic but all of a sudden there are 4 highly rails specific configuration values in here. This should be a separate configuration object which is then included into this one via composition.

lib/rubycritic/generators/html/base.rb
```diff
@@ -10,7 +10,7 @@ def self.erb_template(template_path)
           ERB.new(File.read(File.join(TEMPLATES_DIR, template_path)))
         end
 
-        TEMPLATES_DIR = File.expand_path('../templates', __FILE__)
+        TEMPLATES_DIR = File.expand_path('../app/views/rubycritic', __FILE__)
```
troessner: That is base.rb - this should not be affected by this PR at all.

troessner:
Thanks for the PR! However, there are multiple problems with this as I pointed out in my comments.

Apart from that there are 2 pretty big additional problems:

1.) No tests?! Sorry, but that is a no-go.
2.) This PR shouldn't affect the existing javascript and css files at all. After all, the html layout and the rails specific layout are separate things.

felipegruoso:
Hello @troessner, thanks for your comments.. I'll work a little bit more on it then.

Onumis:
This is looking great! Only lacks some tests now :)

troessner:
@felipegruoso let's get this going again, otherwise we - or rather: you - wasted a lot of energy on nothing :(

felipegruoso:
@Onumis thanks for feedback hehe...
@troessner I'm a little busy lately hehe but I'll finish it soon

troessner:
Closing this one due to inactivity. @felipegruoso feel free to re-open this one when you feel like getting this going again.

PR #148, opened 2016-06-24
Move some wiki pages into /docs

Onumis:
@troessner not sure if we should move the Roadmap here too. It's part a list of links to open issues, part random thoughts.


troessner:
Regarding roadmap, for Reek we use github issues and label them accordingly (at least the breaking changes), how about trying this here as well?
We could have a "roadmap" label for issues for instance.

Onumis:
Sounds good! :+1:

PR #152, opened 2016-07-01
flay 2.8.0, flog 4.4.0, rubocop 0.41.1

nijikon:


nijikon:
@troessner can we push 2.9.2 when this is merged?

PR #155, opened 2016-08-03
Fix Ruby[Cc]ritic case in rake task examples.

Insti:
The Rake task examples used `RubyCritic` as the module name, when it should be `Rubycritic`.

Using the uppercase version would result in the error message:
```
NameError: uninitialized constant RubyCritic
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ Rubycritic::RakeTask.new
 A more sophisticated Rake task that would make use of all available configuration options could look like this:
 
 ```ruby
-RubyCritic::RakeTask.new do |task|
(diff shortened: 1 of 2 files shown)


troessner:
Yeah, that is pretty bad. It actually **should** be RubyCritic and not Rubycritic.
Lets rename this across our codebase.
Wdyt @Onumis / @lucasmazza ?

Onumis:
Yep, should be `RubyCritic`

Insti:
Ok, then the problem is everywhere else. I'll leave that to you.

troessner:
I'll fix this the next days, thanks for bringing this topic up!
Closing this PR.

PR #156, opened 2016-08-03
Eliminate RuboCop offenses to make CI build pass.

Insti:
```
lib/rubycritic/configuration.rb:31:5: C: When using method_missing, define respond_to_missing? and fall back on super.

    def self.method_missing(method, *args, &block) ...

    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

lib/rubycritic/core/analysed_module.rb:30:10: C: Use methods_count.zero? instead of methods_count == 0.

      if methods_count == 0

         ^^^^^^^^^^^^^^^^^^
```
1) Disabled `Style/MethodMissing` check for that method.
2) Changed `methods_count == 0` to `methods_count.zero?`

### DIFF
--- a/lib/rubycritic/configuration.rb
+++ b/lib/rubycritic/configuration.rb
@@ -31,5 +31,9 @@ module Rubycritic
     def self.method_missing(method, *args, &block)
       configuration.public_send(method, *args, &block)
     end
+
+    def self.respond_to_missing?(symbol, include_all = false)
+      configuration.respond_to_missing?(symbol) || super
+    end
   end
 end
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -34,3 +34,9 @@ Style/ClassVars:
 # SupportedStyles: use_perl_names, use_english_names
 Style/SpecialGlobalVars:
   Enabled: false
+
+# Offense count: 1
+Style/MethodMissing:
+  Exclude: 
+    - 'lib/rubycritic/configuration.rb'
+
--- a/.todo.reek
+++ b/.todo.reek
@@ -136,3 +136,6 @@ UncommunicativeParameterName:
 ClassVariable:
   exclude:
   - Rubycritic::SourceControlSystem::Base
+BooleanParameter: 
+  exclude:
+  - Rubycritic::Config#self.respond_to_missing?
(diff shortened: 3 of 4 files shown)


Onumis:
LGTM! Thanks

lib/rubycritic/configuration.rb
```diff
@@ -28,8 +28,10 @@ def self.set(options = {})
       configuration.set(options)
     end
 
+    # rubocop:disable Style/MethodMissing
     def self.method_missing(method, *args, &block)
       configuration.public_send(method, *args, &block)
     end
+    # rubocop:enable Style/MethodMissing
```
troessner: Hmmm, lets rather put this in our configuration file and not in our codebase.

troessner:
One comment and lgtm ;)

Insti:
I tried to resolve the issue by adding the required `respond_to_missing?` code and `super` call which should have fixed it, but there is a Rubocop issue: https://github.com/bbatsov/rubocop/issues/3358
Which means `self.respond_to_missing?` is not detected by Rubocop and the Rubocop error is still raised. 

So I have disabled it again, but this time in the `.rubocop_todo.yml` file.

Once the Rubocop issue is resolved, the exclusion can be removed and everything will be good.

Insti:
Adding `respond_to_missing?` also triggered a reek warning that needed to be excluded.

lib/rubycritic/configuration.rb
```diff
@@ -29,7 +29,15 @@ def self.set(options = {})
     end
 
     def self.method_missing(method, *args, &block)
-      configuration.public_send(method, *args, &block)
+      if configuration
```
troessner: Wait, why is this guard now necessary?

Insti: It was a way to get the `super` into the `method_missing` to appease Rubocop.
It may just be better to disable the cop and put the code back how it was, but I wanted to at least present a version of the code that didn't raise any warnings.

troessner: Yeah, lets rather disable the cop.

Then pls squash your PR to one commit afterwards so we cn merge this bad boy ;)

Insti:
Made changes discussed in https://github.com/whitesmith/rubycritic/pull/156#discussion-diff-73488481 and squashed.

troessner:
Awesome, merged, thanks!

PR #157, opened 2016-08-04
Rename Rubycritic to RubyCritic.

troessner:
Coming from #155 

The command I used for this was:

```Bash
find . -type f | grep -v ".git/" | LC_ALL=C xargs sed -i '' 's/Rubycritic/RubyCritic/g'
```

Let's go through the command:

 ```Bash
find . -type f | grep -v 
```

Necessary to not botch our .git directory. I know there is "-not" and "prune" as well but I never managed to get this to work and every time I have to do "man find" I want to drink myself into oblivion.

```Bash
grep -v ".git/"
```
Mind the "/" at the end. Without it, lib/rubycritic/source_control_systems/git.rb would haven been excluded :)

```Bash
LC_ALL=C
```

Was necessary because of .....well, I honestly do not want to know.

```Bash
sed -i '' 's/Rubycritic/RubyCritic/g'
```

Mind the space between -i and ''. If this isn't there, your bash will blow up.


Onumis:
That's great! I also don't know about `LC_ALL=C`, and I cba to know...
LGTM, merging...

PR #159, opened 2016-08-16
Save json report to file

NickTroy:
Hello!

I need to post-process json data generated in report which is not possible because now json data is printed to console. That's why I added a feature that creates a file 'report.json' in Config.root directory and saves json data there.

Tell me whether this feature needs testing.

With respect, Mykyta Troianov.

### DIFF
--- a/lib/rubycritic/generators/json/simple.rb
+++ b/lib/rubycritic/generators/json/simple.rb
@@ -1,5 +1,6 @@
 require 'json'
 require 'rubycritic/version'
+require 'pathname'
 
 module RubyCritic
   module Generator
@@ -8,6 +9,8 @@ module RubyCritic
         def initialize(analysed_modules)
           @analysed_modules = analysed_modules
         end
+
+        FILE_NAME = 'report.json'.freeze
 
         def render
           JSON.dump(data)
@@ -23,6 +26,14 @@ module RubyCritic
             analysed_modules: @analysed_modules,
             score: @analysed_modules.score
           }
+        end
+
+        def file_directory
+          @file_directory ||= Pathname.new(Config.root)
+        end
+
+        def file_pathname
+          Pathname.new(file_directory).join FILE_NAME
         end
       end
     end
--- a/lib/rubycritic/generators/json_report.rb
+++ b/lib/rubycritic/generators/json_report.rb
@@ -8,7 +8,10 @@ module RubyCritic
       end
 
       def generate_report
-        print generator.render
+        FileUtils.mkdir_p(generator.file_directory)
+        File.open(generator.file_pathname, 'w+') do |file|
+          file.write(generator.render)
+        end
       end
 
       private
--- a/test/lib/rubycritic/generators/json_report_test.rb
+++ b/test/lib/rubycritic/generators/json_report_test.rb
@@ -1,0 +1,36 @@
+require 'test_helper'
+require 'rubycritic/core/analysed_modules_collection'
+require 'rubycritic/generators/json_report'
+require 'json'
+require 'fakefs/safe'
+
+describe RubyCritic::Generator::JsonReport do
+  describe '#generate_report' do
+    before(:each) do
+      FakeFS.activate!
+      create_analysed_modules_collection
+      generate_report
+    end
+
+    after(:each) { FakeFS.deactivate! }
+
+    it 'creates a report.json file' do
+      assert File.file?('test/samples/report.json'), 'expected report.json file to be created'
+    end
+
+    it 'report file has data inside' do
+      data = File.read('test/samples/report.json')
+      assert data != '', 'expected report file not to be empty'
+    end
+  end
+
+  def create_analysed_modules_collection
+    @analysed_modules_collection = RubyCritic::AnalysedModulesCollection.new('test/samples/')
+    RubyCritic::Config.root = 'test/samples'
+  end
+
+  def generate_report
+    report = RubyCritic::Generator::JsonReport.new(@analysed_modules_collection)
+    report.generate_report
+  end
+end
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ rdoc
 spec/reports
 test/tmp
 test/version_tmp
+test/samples/report.json
 tmp
 .idea/
 .ruby-gemset
--- a/rubycritic.gemspec
+++ b/rubycritic.gemspec
@@ -32,6 +32,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'aruba'
   spec.add_development_dependency 'bundler', '~> 1.3'
   spec.add_development_dependency 'cucumber'
+  spec.add_development_dependency 'fakefs'
   spec.add_development_dependency 'rake'
   spec.add_development_dependency 'minitest', '~> 5.3'
   spec.add_development_dependency 'mocha', '~> 1.1'


lib/rubycritic/generators/json/simple.rb
```diff
@@ -24,6 +25,18 @@ def data
...
+        def file_directory
+          @file_directory ||= Pathname.new(Config.root)
+        end
+
+        def file_name
+          'report.json'
+        end
+
+        def file_pathname
+          File.join(file_directory, file_name)
```
troessner: Why use File here and not Pathname? Also I believe you can merge `file_name` and `file_pathname`, this can be one method.

lib/rubycritic/generators/json/simple.rb
```diff
@@ -24,6 +25,18 @@ def data
             score: @analysed_modules.score
           }
         end
+
+        def file_directory
+          @file_directory ||= Pathname.new(Config.root)
+        end
+
+        def file_name
+          'report.json'
```
troessner: Let's make this a constant.

lib/rubycritic/generators/json/simple.rb
```diff
@@ -24,6 +25,18 @@ def data
             score: @analysed_modules.score
           }
         end
+
+        def file_directory
+          @file_directory ||= Pathname.new(Config.root)
```
troessner: I think this should be configurable and not kind of hard-coded.

troessner:
I like the proposal but I believe this needs some polishing before this can go live.
I see 2 big issues:

1.) No specs
2.) The path directory should be configurable because otherwise this will pollute the project directory which will alienate users

Apart from that please see my other comments.

Let's get this going!

lib/rubycritic/generators/json/simple.rb
NickTroy: Merged #file_name and #file_pathname to #file_pathname. I used File.join because it looks simpler to me, can substitute it with  Pathname.new(file_directory, FILE_NAME), should I?

lib/rubycritic/generators/json/simple.rb
NickTroy: Made it a constant.

lib/rubycritic/generators/json/simple.rb
NickTroy: At the moment user can configure saving directory by using -p "path/to/directory/", this feature is still working.

NickTroy:
Completely agree with you about polishing. But will need your advise on:
1) What exactly should be tested in this feature? As I see it now this should be covered: creating file, checking correctness of directory, checking data inside file.
Did I understand you correct about configurable file directory?

lib/rubycritic/generators/json/simple.rb
troessner: >> At the moment user can configure saving directory by using -p "path/to/directory/", this feature is still working.

Good point, then please just ignore my comment :)

troessner:
>> As I see it now this should be covered: creating file, checking correctness of directory, checking data inside file.

We don't have to be to pedantic about it but I think we should at least check that a non-empty file was generated.

>> Did I understand you correct about configurable file directory?

Yeah, see the comment above, you can ignore this ;)

NickTroy:
> We don't have to be to pedantic about it but I think we should at least check that a non-empty file was generated.
I will add a test that checks for non-empty file report.json in concrete folder.

NickTroy:
Added a test to check whether file exists and whether it's not empty. Advise on adding any other tests, if they are needed.

test/lib/rubycritic/generators/json_report_test.rb
```diff
@@ -0,0 +1,40 @@
+require 'test_helper'
+require 'rubycritic/core/analysed_modules_collection'
+require 'rubycritic/generators/json_report'
+require 'json'
+
+describe RubyCritic::Generator::JsonReport do
+  describe '#generate_report' do
+    before do
+      FileUtils.rm('test/samples/report.json') if File.file?('test/samples/report.json')
```
troessner: Judging from experience I'd say that this will turn into a flickering test one day since you're operating on the real filesystem. How about using [fakefs](https://github.com/fakefs/fakefs) to defuse this bomb right from the start?

troessner: This will also allow us to drop the line above

test/lib/rubycritic/generators/json_report_test.rb
```diff
@@ -0,0 +1,40 @@
...
+      create_analysed_modules_collection
+      generate_report
```
troessner: I find it confusing that the examples below just contain "assert". I think we should move the report generation into the examples. Using a fake file system like suggested above will also make this easier since you don't have to clean up afterwards. Meaning that you can just move "generate_report" to each examples with duplicating the set up everywhere.

lib/rubycritic/generators/json_report.rb
```diff
@@ -8,7 +8,10 @@ def initialize(analysed_modules)
       end
 
       def generate_report
-        print generator.render
+        FileUtils.mkdir_p(generator.file_directory)
```
troessner: We can and should use [Pathname](https://ruby-doc.org/stdlib-2.1.0/libdoc/pathname/rdoc/Pathname.html#method-i-mkdir) here.

lib/rubycritic/generators/json_report.rb
```diff
@@ -8,7 +8,10 @@ def initialize(analysed_modules)
...
+        File.open(generator.file_pathname, 'w+') do |file|
```
troessner: And here.

lib/rubycritic/generators/json/simple.rb
```diff
@@ -24,6 +27,14 @@ def data
...
+        end
+
+        def file_pathname
+          File.join(file_directory, FILE_NAME)
```
troessner: Let's re-write this to

```Ruby
Pathname.new(file_directory).join FILE_NAME
```

lib/rubycritic/generators/json_report.rb
NickTroy: This would look more complicated, because #mkdir throws error if folder already exists.
`Pathname.new(generator.file_directory).mkdir` unless File.directory? generator.file_directory

lib/rubycritic/generators/json_report.rb
NickTroy: Same as above.

test/lib/rubycritic/generators/json_report_test.rb
NickTroy: Reworked tests to use FakeFS.

test/lib/rubycritic/generators/json_report_test.rb
NickTroy: In recent change I added report is generated before each test using FakeFS.

NickTroy:
Tell me, if you think that we should use Pathname anyway, will rework it asap.

lib/rubycritic/generators/json_report.rb
troessner: Alright, fair enough.

troessner:
Looking good now, merged, thanks a lot!

NickTroy:
Was a great experience for me! Thank you as well!

NickTroy:
One final question, when this feature will be published to rubygems?

troessner:
>> One final question, when this feature will be published to rubygems?

2 minutes ago:

>> Tagged v2.9.3.
Pushed git commits and tags.
Pushed rubycritic 2.9.3 to rubygems.org

😆

NickTroy:
Thanks again!)

PR #163, opened 2016-09-14
Update reek version and flog/flay/rubocop

bglusman:
Reek had a bug, https://github.com/troessner/reek/issues/960 which is now fixed, but Rubycritic is using previous version.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 source 'https://rubygems.org'
 
 # Specify your gem's dependencies in rubycritic.gemspec
--- a/Rakefile
+++ b/Rakefile
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 require 'bundler/gem_tasks'
 require 'rake/testtask'
 require 'rubocop/rake_task'
--- a/bin/rubycritic
+++ b/bin/rubycritic
@@ -1,4 +1,5 @@
 #!/usr/bin/env ruby
+# frozen_string_literal: true
 
 # Always look in the lib directory of this gem
 # first when searching the load path
--- a/features/step_definitions/rake_task_steps.rb
+++ b/features/step_definitions/rake_task_steps.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 When(/^I run rake (\w*) with:$/) do |name, task_def|
   rake(name, task_def)
 end
--- a/features/step_definitions/rubycritic_steps.rb
+++ b/features/step_definitions/rubycritic_steps.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 When(/^I run rubycritic (.*)$/) do |args|
   rubycritic(args)
 end
--- a/features/step_definitions/sample_file_steps.rb
+++ b/features/step_definitions/sample_file_steps.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 Given(/^the smelly file 'smelly.rb'/) do
   contents = <<-EOS.strip_heredoc
     class AllTheMethods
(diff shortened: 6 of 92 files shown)


bglusman:
Thought probably more correct thing is to not specify the patch version in reek or several other gems...  though weirdly that gave warnings in travis/rubocop...  reverted back to this version but curious to hear someone's take on the warnings/non-patch version, available on my fork at https://github.com/bglusman/rubycritic/tree/less_strict_versionig

troessner:
4.1.1. Is quite old and we added some awesome new detectors in the meantime, so lets rather make this 4.4.1 :)

bglusman:
Oh, good call, I misread that as newest version... updating PR...

bglusman:
Weirdly rubocop errors for me locally with last changes, but passes on Travis... on 2.3.0, same version as Travis. Kind of understand why it would happen on pre-2.3 versions where frozen pragma has no effect, but weird that it wants it in 2.3 also.

Onumis:
Wow! So many changes! Thanks a lot for the effort!

Maybe updating rubocop version can help. This looks fine, but I'll want to look into this myself :)

troessner:
I'm sorry, but I'm afraid this PR is not really reviewable in its current form.
We started off with an innocent dependency update and now we are having multiple, completely unrelated updates in one PR.
I don't feel confident in reviewing this at all especially since you have [non-trivial changes](https://github.com/whitesmith/rubycritic/pull/163/files#diff-d326961c8359d73a279d6da03d3c44e6L25) in it as well.
Can you please split this up into multiple PRs which are atomic (so they have one purpose and one purpose only)?

bglusman:
They're not at all unrelated, they're exclusively related to the update and getting tests/rubocop warnings (which fail the test suite) to pass, though there are two sperate judgment calls I'm happy to discuss or reverse.  

1) There only non trivial change you point out is switching from string mutation to array based accumulation, which seems more idiomatic to me, and I only did that because the alternative  was adding an explicit disable for the rubocop frozen string literal cop, and fixing the issue seemed preferable to disabling the warning.
2) as mentioned above, it seemed more correct to not lock down precise gem patch versions where it could be avoided, so aside from reek I made a similar change to flog and flay versions.

Happy to reverse or consider alternatives to either or both of them, and I may be missing some setting to make rubocop/test suite behave identically locally to how it does on Travis, but I ran and based decisions on 2.3.0 ruby behavior with bundle exec take, but simply splitting up the changes will result in a PR that's not passing, unless I'm misunderstanding you, or like I said perhaps running tests wrong locally, but I think the Travis fail history above makes reasonably clear most of the changes are necessary for the test suite.  Thanks and sorry it grew in scope as a result, the initial version that only went to version 4.1.1 is still an option of course.

troessner:
I see your point @bglusman and I'm convinced.
I did review the PR just now and this looks good to me.
One last thing to do: Please squash to one commit and then this is good to merge. We also need to name that one commit properly - "Update Reek, flog, flay and update to latest rubocop requirements." or similar might be appropriate :)

Onumis:
Can we also update rubocop to v0.42.0? Just so we've everything covered :)

bglusman:
I think that rebased version should address all issues, thanks!

troessner:
Great job, thanks a lot, merged!

troessner:
Just released RC 2.9.4 with our changes!

guilhermesimoes:
I guess I'm a little late, but I just you wanted to explain my reasoning for all the gems being locked to specific versions.

Flog, Flay and Reek are all being used in ways that were not meant to be. They don't have clear APIs for accessing their analysis results and RubyCritic accesses internal structures of those gems to get those results. I couldn't be sure that a patch version wouldn't change those internals and break RubyCritic and that's why I locked those versions.
(I believe this no longer applies to Reek though, given that @troessner has been working on both gems 😄 )

I've been following (not very closely) RubyCritic's maintenance and I've noticed that the gem versions being locked were causing you guys a lot of entropy. So I totally understand why you made this change.

But one day a user may report a weird bug difficult to track down and that's because he will have installed a slightly different version of Flog or Flay. Or maybe not. In any case, I just wanted to let you guys know 😄

PR #165, opened 2016-10-06
Support Perforce source control system for Churn metric

Rataah:
This pull request adds the support of Perforce as source control for the Churn metric.

When we request informations on a file, the Perforce client retrieves the statistic from the server, which is very slow.
For better performances, a cache is built with all ruby files present in the current directory. 

A new unit test is created to ensure the class respond to the "BasicInterface".

Let me know if something is missing.

**note**: I have developed this on Windows. 2 unit tests fail on my computer but it should be OK
[unit_tests_fail.txt](https://github.com/whitesmith/rubycritic/files/514254/unit_tests_fail.txt)

### DIFF
--- a/lib/rubycritic/source_control_systems/base.rb
+++ b/lib/rubycritic/source_control_systems/base.rb
@@ -34,3 +34,4 @@ end
 require 'rubycritic/source_control_systems/double'
 require 'rubycritic/source_control_systems/git'
 require 'rubycritic/source_control_systems/mercurial'
+require 'rubycritic/source_control_systems/perforce'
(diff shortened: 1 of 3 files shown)


troessner:
Thanks for this PR!
The biggest problem here is that there are almost no tests.
If you add extensive test coverage I'm happy to review this bad boy ;)

Rataah:
Hi,

Here is some new unit tests to have a better coverage.
I rewrote some parts of my code to be cleaner

troessner:
I have no idea about Perforce but given that there are now meaningful tests I'm willing to merge this one.
@Onumis wdyt?
@Rataah can you please squash your commits to one? (pre-condition to merge)

Onumis:
LGTM.
We'll need to update the README to mention that this will only work with the env flag `P4CLIENT`.
I noticed that the README is quite outdated regarding supported source control systems, so I'll update it myself :)

Rataah:
I squashed all my commits into a single one.
Cheers =)

troessner:
Great job! Merged!

PR #167, opened 2016-10-16
GPL license conformance

steakknife:
Using any GPL-licensed gem in a non-GPL gem, project (or dependency) is often problematic, possibly a GPL license violation but IANAL. Using [license_finder](https://github.com/pivotal/LicenseFinder), the only rubycritic dep. out of license compatibility is [colorize](https://github.com/fazibear/colorize). This PR is ~~a quick hack~~ refactored to use [rainbow](https://github.com/sickill/rainbow) instead, since terminal colorizing is trivial and there are [many options](https://www.ruby-toolbox.com/categories/Terminal_Coloring).

### DIFF
--- a/lib/rubycritic/generators/text/list.rb
+++ b/lib/rubycritic/generators/text/list.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-require 'colorize'
+require 'rainbow'
 
 module RubyCritic
   module Generator
--- a/lib/rubycritic/generators/text/templates/list.erb
+++ b/lib/rubycritic/generators/text/templates/list.erb
@@ -1,5 +1,5 @@
-<%= @analysed_module.name.colorize(color) %>:
-  Rating:       <%= @analysed_module.rating.to_s.colorize(color) %>
+<%= Rainbow(@analysed_module.name.to_s).foreground(color) %>:
+  Rating:       <%= Rainbow(@analysed_module.rating.to_s).foreground(color) %>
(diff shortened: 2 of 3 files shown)


lib/rubycritic/generators/text/list.rb
```diff
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
-require 'colorize'
+require 'rainbow/ext/string'
+
+class String
+  alias colorize color
```
troessner: Can we solve this problem without monkey patching?

troessner:
Looks better!

Left to do:

* replace "send" with "public_send"
* you didnt update the gemspec ;)
* make travis happy
* squash to one commit

steakknife:
Github automagically-closes empty PRs

steakknife:
- `send` -> `foreground` (safer)
- `rubycritic.gemspec` was already updated
- travis: build broke between last commit on master and now (as per c1c54545dcecd4701a737b9e9f823518762066c0), probably a dependency leak on rubocop because of lack of checked-in `Gemfile.lock`
- rebased

lib/rubycritic/generators/text/list.rb
steakknife: We have the (refactoring) technology, we can build him (per Working Effectively with Legacy Code).

steakknife:
This should be rebased onto HEAD after merging #168 (which Github allows maintainers to do now)

PR #168, opened 2016-10-16
Fix travis (merge me first)

steakknife:
This should work. 😎

PR #170, opened 2016-10-16
Remove Gemfile.lock from version control

Onumis:
In https://github.com/whitesmith/rubycritic/pull/168 we added Gemfile.lock to the repository, I believe this was a mistake, and we should remove it.

Reasoning: http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/


troessner:
Yes, my bad, I didnt pay enough attention, will merge when green.

Onumis:
Nah, it was me that pulled the trigger :P

steakknife:
The actual issue though is that when dependencies change, a travis-ci build needs to be triggered.  Otherwise, the build looks like it "passes" when it's broken.

PR #171, opened 2016-10-16
Set required ruby version to 2.1

Onumis:
We've removed tests on Ruby 2.0 [some time ago](https://github.com/whitesmith/rubycritic/commit/02d1451e03a54dac95df86a183708eb1fdd42c95)

I think we should also explicitly say we don't support it.


troessner:
Good catch!

PR #174, opened 2016-10-21
Add and increase a bit the turboThreshold of Highcharts. 

victormartins:
Hello guys, first of all thanks for an amazing work, really enjoying this gem.

I've just added it to a project with a few years and at first the graphic plotting the churn and complexity didn't show up. The error pointed me to the right solution, it was the turboThreshold property, I had to raise it from 1000 to 1500.
I've decided to PR this, cause it may help others and at least it will leave a placeholder to make it easier to find and change.

### DIFF
--- a/lib/rubycritic/generators/html/assets/javascripts/application.js
+++ b/lib/rubycritic/generators/html/assets/javascripts/application.js
@@ -80,6 +80,9 @@ $("#js-chart-container").highcharts({
         endOnTick: true
     },
     plotOptions: {
+        series: {
+            turboThreshold: 1500
+        },
         scatter: {
             marker: {
                 radius: 5,


Onumis:
Hi! Maybe we can just set it to `0` and have it disabled?
I'm not sure on the impact this can have on the load times of the page, but I guess that's not of much importance.

victormartins:
@Onumis  Zero is an option, but it leaves the door open for "break" if someone has loads of data.
On my example I felt no noticeable impact on the page load time.

troessner:
I'm for merging this one. @Onumis ok for you?

Onumis:
Yeah!
Thanks @victormartins

PR #175, opened 2016-10-31
Respect Reeks config file search

mereghost:
Rubycritic was using a subclass of Reek::Examiner. The examiner does not search (or load) the config files by itself, so we need to allow the default behavior used by the `Reek::CLI::Application` which is passing to the examiner a `Reek::Configuration::AppConfiguration.from_path`  with either a `path` to a .reek file or a `nil` (which will trigger the searching for .reek files).

I'm not that all happy with the way I got to make a test for this, but it works and should prevent regressions.

### DIFF
--- a/.todo.reek
+++ b/.todo.reek
@@ -133,9 +133,10 @@ UncommunicativeParameterName:
   - RubyCritic::Location#to_json
   - RubyCritic::Rating#to_json
   - RubyCritic::Smell#to_json
+  - Theon # This is needed for the tests to pass
 ClassVariable:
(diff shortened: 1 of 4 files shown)


mereghost:
For some reason I somehow broke it. =/

troessner:
Very nice! Can you please squash to one commit? I'll then happily merge and release this one.

mereghost:
Squashed. =)

troessner:
Great job, thanks!
I can release it tmr unless somebody is faster.

PR #176, opened 2016-11-01
Bump version to 3.0.0.

troessner:


troessner:
cc @mereghost

Onumis:
Why not `2.10`?
There are no substantial / backwards incompatible changes right?

troessner:
>> There are no substantial / backwards incompatible changes right?

We're dropping 2.0 support, which I'd call an incompatible change for sure :)

Onumis:
Ah yes, you're right!

Onumis:
Published :tada:

troessner:
Awesome!

PR #179, opened 2016-11-03
Spec: Fix copy-pasted titles of "it" cases

olleolleolle:
In order to keep the test readable, this PR changes the names of the test cases in ConsoleReport's spec.

It seems some of them were copy-paste oopsies.

Great to have specs, though!


troessner:
Good catch, merged, thanks!

PR #183, opened 2016-11-14
Fix some ruby warnings thrown on rake test

tejasbubane:
Running `rake` gives following warnings: https://travis-ci.org/whitesmith/rubycritic/jobs/174016330#L237-L239

Upgrade `parser` to `2.3.1.4`.
One warning (the last one) still left which comes from `parser`. Hopefully that is fixed in later versions.
One way is to disable warnings completely on `rake test`, but I don't think that will be appropriate.


troessner:
Merged, thanks for that!

Issue #161, opened 2016-08-22
CI mode does not run only on the diff from master


thedrow:
I don't really understand what rubycritic does when CI mode is on but it doesn't simply run on the files that have changed since the branching from master.
Is there a way to make rubycritic do so?

thedrow:
```bash
DIFF_LIST=$(git diff --name-only $CIRCLE_SHA1 $(git merge-base $CIRCLE_SHA1 master) | grep .rb)
  if [ ! -z "$DIFF_LIST" ]; then
    bundle exec rubycritic --format console -m $DIFF_LIST
  else
    echo "No changes detected. Skipping rubycritic..."
  fi
```
does the trick on CircleCI. Maybe we should document this somehow?

Onumis:
The CI mode runs only against the last commit. Doing it against the full diff of a branch is indeed very helpful. 
Gonna add this as a task.
Thanks for the CircleCI snippet, ideally, RubyCritic should handle that.

thedrow:
I don't understand the difference between the CI mode and normal mode?
Are you able to visualize code quality over time?

PR #191, opened 2016-11-29
Compare only the modified files between two branches

HemanthMudalaiah:
Trying to solve [https://github.com/whitesmith/rubycritic/issues/161](https://github.com/whitesmith/rubycritic/issues/161)

Introducing -b option to compare files between two branches and another -t option is to set threshold for score difference between two branches.

Creating a separate build view which will have only the modified files in the code_index page, and links to two compared branches.
screen shot 2016-12-15 at 4 49 38 pm

### DIFF
--- a/lib/rubycritic/cli/options.rb
+++ b/lib/rubycritic/cli/options.rb
@@ -19,6 +19,17 @@ module RubyCritic
             @root = path
           end
 
+          opts.on('-b', '--branch BRANCH', 'Set branch to compare') do |branch|
+            self.base_branch = String(branch)
+            set_current_branch
+            self.mode = :compare_branches
+          end
+
+          opts.on('-t', '--maximum-decrease [MAX_DECREASE]',
+                  'Set a threshold for score difference between two branches (works only with -b)') do |threshold_score|
+            self.threshold_score = Integer(threshold_score)
+          end
+
           opts.on(
             '-f', '--format [FORMAT]',
             [:html, :json, :console],
@@ -34,7 +45,10 @@ module RubyCritic
             self.minimum_score = Integer(min_score)
           end
 
-          opts.on('-m', '--mode-ci', 'Use CI mode (faster, but only analyses last commit)') do
+          opts.on('-m', '--mode-ci [BASE_BRANCH]',
+                  'Use CI mode (faster, analyses diffs w.r.t base_branch (default: master))') do |branch|
+            self.base_branch = branch || 'master'
+            set_current_branch
             self.mode = :ci
           end
 
@@ -71,20 +85,28 @@ module RubyCritic
           suppress_ratings: suppress_ratings,
           help_text: parser.help,
           minimum_score: minimum_score || 0,
-          no_browser: no_browser
+          no_browser: no_browser,
+          base_branch: base_branch,
+          feature_branch: feature_branch,
+          threshold_score: threshold_score || 0
         }
       end
 
       private
 
       attr_accessor :mode, :root, :format, :deduplicate_symlinks,
-                    :suppress_ratings, :minimum_score, :no_browser, :parser
+                    :suppress_ratings, :minimum_score, :no_browser,
+                    :parser, :base_branch, :feature_branch, :threshold_score
       def paths
         if @argv.empty?
           ['.']
         else
           @argv
         end
+      end
+
+      def set_current_branch
+        self.feature_branch = SourceControlSystem::Git.current_branch
       end
     end
   end
--- a/lib/rubycritic/commands/compare.rb
+++ b/lib/rubycritic/commands/compare.rb
@@ -39,6 +39,7 @@ module RubyCritic
       # keep track of the number of builds and
       # use this build number to create separate directory for each build
       def update_build_number
+        FileUtils.mkdir_p(Config.root) unless File.directory?(Config.root)
         build_file_location = "#{Config.root}/build_count.txt"
         File.new(build_file_location, 'a') unless File.exist?(build_file_location)
         @build_number = File.open(build_file_location).readlines.first.to_i + 1
--- a/lib/rubycritic/commands/compare.rb
+++ b/lib/rubycritic/commands/compare.rb
@@ -37,7 +37,7 @@ module RubyCritic
       end
 
       # keep track of the number of builds and
-      # use this build number to create seperate directory for each build
+      # use this build number to create separate directory for each build
       def update_build_number
         build_file_location = "#{Config.root}/build_count.txt"
         File.new(build_file_location, 'a') unless File.exist?(build_file_location)
--- a/lib/rubycritic/commands/compare.rb
+++ b/lib/rubycritic/commands/compare.rb
@@ -27,10 +27,11 @@ module RubyCritic
       def compare_branches
         update_build_number
         set_root_paths
+        original_no_browser_config = Config.no_browser
         Config.no_browser = true
         analyse_branch(:base_branch)
         analyse_branch(:feature_branch)
-        Config.no_browser = false
+        Config.no_browser = original_no_browser_config
         analyse_modified_files
         compare_code_quality
       end
@@ -38,7 +39,7 @@ module RubyCritic
       # keep track of the number of builds and
       # use this build number to create seperate directory for each build
       def update_build_number
-        build_file_location = '/tmp/build_count.txt'
+        build_file_location = "#{Config.root}/build_count.txt"
         File.new(build_file_location, 'a') unless File.exist?(build_file_location)
         @build_number = File.open(build_file_location).readlines.first.to_i + 1
         File.write(build_file_location, @build_number)
@@ -90,11 +91,11 @@ module RubyCritic
       end
 
       def branch_directory(branch)
-        "tmp/rubycritic/compare/#{Config.send(branch)}"
+        "#{Config.root}/compare/#{Config.send(branch)}"
       end
 
       def build_directory
-        "tmp/rubycritic/compare/builds/build_#{@build_number}"
+        "#{Config.root}/compare/builds/build_#{@build_number}"
       end
 
       # create a txt file with the branch score details
--- a/lib/rubycritic/commands/compare.rb
+++ b/lib/rubycritic/commands/compare.rb
@@ -1,0 +1,115 @@
+# frozen_string_literal: true
+require 'rubycritic/source_control_systems/base'
+require 'rubycritic/analysers_runner'
+require 'rubycritic/revision_comparator'
+require 'rubycritic/reporter'
+require 'rubycritic/commands/base'
+require 'rubycritic/commands/default'
+
+module RubyCritic
+  module Command
+    class Compare < Default
+      def initialize(options)
+        super
+        @build_number = 0
+      end
+
+      def execute
+        compare_branches
+        status_reporter.score = Config.feature_branch_score
+        status_reporter
+      end
+
+      private
+
+      attr_reader :paths, :status_reporter
+
+      def compare_branches
+        update_build_number
+        set_root_paths
+        Config.no_browser = true
+        analyse_branch(:base_branch)
+        analyse_branch(:feature_branch)
+        Config.no_browser = false
+        analyse_modified_files
+        compare_code_quality
+      end
+
+      # keep track of the number of builds and
+      # use this build number to create seperate directory for each build
+      def update_build_number
+        build_file_location = '/tmp/build_count.txt'
+        File.new(build_file_location, 'a') unless File.exist?(build_file_location)
+        @build_number = File.open(build_file_location).readlines.first.to_i + 1
+        File.write(build_file_location, @build_number)
+      end
+
+      def set_root_paths
+        Config.base_root_directory = Pathname.new(branch_directory(:base_branch))
+        Config.feature_root_directory = Pathname.new(branch_directory(:feature_branch))
+        Config.build_root_directory = Pathname.new(build_directory)
+      end
+
+      # switch branch and analyse files
+      def analyse_branch(branch)
+        SourceControlSystem::Git.switch_branch(Config.send(branch))
+        critic = critique(branch)
+        Config.send(:"#{branch}_score=", critic.score)
+        Config.root = branch_directory(branch)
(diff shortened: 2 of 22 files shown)


troessner:
Wow, that is a pretty big, promising PR! Unfortunately I'm swamped at the moment though unfortunately and won't be able to review this any time soon. @Onumis @lucasmazza can you take this over?

Onumis:
Weekdays are tough, but I think I can have a look at this later today ;)
This is an awesome feature, looking forward to get this merged!

thedrow:
@Onumis This looks great. Any change to get this merged?

lib/rubycritic/commands/compare.rb
```diff
@@ -0,0 +1,115 @@
...
+        SourceControlSystem::Git.switch_branch(Config.send(branch))
+        critic = critique(branch)
+        Config.send(:"#{branch}_score=", critic.score)
+        Config.root = branch_directory(branch)
+        report(critic)
+      end
+
+      # generate report only for modified files
+      def analyse_modified_files
+        Config.no_browser = false
```
thedrow: Why is this hardcoded?

thedrow:
Now that I think about it, it'd be more accurate to compare branches more granularly using the lines that have changed in each patch.
It's more work to implement though so we can open an issue about this idea after we merge this PR.

lib/rubycritic/commands/compare.rb
```diff
@@ -0,0 +1,115 @@
...
+        defected_modules = Config.feature_branch_collection.where(SourceControlSystem::Git.modified_files)
```
Onumis: I think you meant `detected`

lib/rubycritic/generators/html/templates/layouts/application.html.erb
```diff
@@ -19,6 +19,13 @@
     <header class="navbar navbar-default navbar-fixed-top">
       <a href="#menu-toggle" class="btn btn-default hidden-lg visible-sm-* hidden-md visible-xs-* pull-left" id="menu-toggle"><i class="fa fa-bars" aria-hidden="true"></i></a>
       <a href="<%= file_path('overview.html') %>"><img src="<%= image_path('logo.png') %>" title="Ruby Critic Logo" width="55"><span class="logo">RUBYCRITIC</span></a>
+      <% if Config.compare_branches_mode? %>
+        <ul class="nav navbar-nav navbar-right">
+          <a href="<%= code_index_path("#{Config.base_root_directory}") %>"><span class="branch"><%= Config.base_branch %></span></a>
+          <a href="<%= code_index_path("#{Config.feature_root_directory}") %>"><span class="branch"><%= Config.feature_branch %></span></a>
+          <a href="<%= code_index_path("#{Config.build_root_directory}") %>"><span class="branch">Build</span></a>
```
Onumis: Ideally, these links would change the "branch" but stay on the same page.

lib/rubycritic/commands/compare.rb
```diff
@@ -0,0 +1,115 @@
...
+        build_file_location = '/tmp/build_count.txt'
+        File.new(build_file_location, 'a') unless File.exist?(build_file_location)
+        @build_number = File.open(build_file_location).readlines.first.to_i + 1
+        File.write(build_file_location, @build_number)
+      end
+
+      def set_root_paths
+        Config.base_root_directory = Pathname.new(branch_directory(:base_branch))
+        Config.feature_root_directory = Pathname.new(branch_directory(:feature_branch))
+        Config.build_root_directory = Pathname.new(build_directory)
```
Onumis: If I understand correctly, this "build" folder contains the rubycritic's results for the files changed between the two branches, correct? Perhaps we could use a better name :thinking:

lib/rubycritic/cli/options.rb
```diff
@@ -19,6 +19,16 @@ def parse
             @root = path
           end
 
+          opts.on('-b', '--branch BASE_BRANCH,FEATURE_BRANCH', 'Set branches to compare') do |branches|
+            self.base_branch, self.feature_branch = branches.split(',').map { |branch| String(branch) }
```
Onumis: what do you think of raising an error if the two branch names are the same?
Since it will obviously not produce any useful result, and will run the analysers twice

Onumis: Also, I think we can use only one branch name, and assume the currently checked-out branch is the "feature branch".
We can also assume a default "base branch", like master, since I guess it's the common case, and it'll make using this feature as simple as: `rubycritic -b`. Thoughts?

lib/rubycritic/cli/options.rb
```diff
@@ -19,6 +19,16 @@ def parse
...
+            self.mode = :compare_branches
+          end
+
+          opts.on('-t', '--threshold-score [THRESHOLD_SCORE]',
+                  'Set a threshold for score difference between two branches (works only with -b)') do |threshold_score|
+            self.threshold_score = Integer(threshold_score)
```
Onumis: I'm not sure I understand the usefulness of this. Isn't it similar to the `--minimum-score` flag?

lib/rubycritic/commands/compare.rb
Onumis: Strange, indeed.

Onumis:
@thedrow 

> it'd be more accurate to compare branches more granularly using the lines that have changed in each patch.

Rubycritic is more oriented towards a whole file. 
But I like the idea, we better put it as a future task.

lib/rubycritic/cli/options.rb
thedrow: It's like codecov's patch difference. If the patch contains too many new code smells we reject it.

lib/rubycritic/cli/options.rb
thedrow: I completely agree.

lib/rubycritic/commands/compare.rb
HemanthMudalaiah: @Onumis ya only the files that were modified in the feature branch, i'll change it to modified modules.

lib/rubycritic/cli/options.rb
HemanthMudalaiah: @Onumis This sounds good, also can we merge this with CI mode since we're comparing only the modified files between two branches. So that we can run ``rubycritic --mode-ci`` and set the currently checked-out branch as the "feature branch" and master will be the "base branch"

lib/rubycritic/generators/html/templates/layouts/application.html.erb
HemanthMudalaiah: @Onumis These are the links that appear in header, which will help you navigate and can you be more specific on ``stay on the same page``

lib/rubycritic/commands/compare.rb
HemanthMudalaiah: @Onumis Ya the build folder contains the rubycritic's results.

Onumis: The thing is "build" is a very generic term. What about "diff" or "files_changed"?

tejasbubane: This feature was developed with the existing `CI mode` in mind. Hoping this would eventually  become `CI mode` (since this is what CI mode is expected to do if I am not wrong).
Hence the term `build`.

lib/rubycritic/cli/options.rb
tejasbubane: Build will fail if the decrease in score is more than this value. These were inspired from `Coveralls` settings:

screen shot 2016-12-15 at 10 49 19 pm

rubycritic already had `--minimum-score`, so we added the decrease threshold. 
Renamed it to `--maximum-decrease` for clarity.

lib/rubycritic/generators/html/templates/layouts/application.html.erb
HemanthMudalaiah: @Onumis This change has been pushed.

HemanthMudalaiah:
@Onumis @tejasbubane Requested changes are pushed.

lib/rubycritic/commands/compare.rb
```diff
@@ -0,0 +1,115 @@
...
+        print("Difference: #{(Config.base_branch_score - Config.feature_branch_score).abs}\n")
+        abort('The score difference between the two branches is over the threshold.')
+      end
+
+      def mark_build_fail?
+        Config.threshold_score > 0 && threshold_reached?
+      end
+
+      def threshold_reached?
+        (Config.base_branch_score - Config.feature_branch_score).abs > Config.threshold_score
```
danmarcab: Just tried and this will make it fail even when you improve the score, I think you should remove the abs

HemanthMudalaiah: @danmarcab Thanks for pointing it out, i've pushed the changes.

vijayanandnandam:
@Onumis @troessner any update on when this can be merged?

Onumis:
@vijayanandnandam the only thing I find missing is what we talked about here: https://github.com/whitesmith/rubycritic/pull/191/files#r92583493

(and squashing, but we can do that directly when merging now..)

@HemanthMudalaiah, would love to get this merged :)

HemanthMudalaiah:
@Onumis @vijayanandnandam  All the requested changes have been pushed :)

Onumis:
@HemanthMudalaiah great!
I see that the `--mode-ci` is working correctly, but the `--branch` does not seem like so:

- I can run `bin/rubycritic -b` with 1 argument, and no error is raised
- It is changing to the "base branch" while analysing, but does not restore to the branch checked out before the run 
- I think it should only accept 1 argument, the branch to compare with (using the current branch as the "feature branch", this way is more in line with `--mode-ci`

what do you think?

HemanthMudalaiah:
@Onumis `rubycritic -b` now accepts 1 argument ( the branch which has to compare with the current branch)

lib/rubycritic/commands/compare.rb
```diff
@@ -0,0 +1,115 @@
...
+
+      attr_reader :paths, :status_reporter
+
+      def compare_branches
+        update_build_number
+        set_root_paths
+        Config.no_browser = true
+        analyse_branch(:base_branch)
+        analyse_branch(:feature_branch)
+        Config.no_browser = false
```
Onumis: this will override the options passed by arguments.

Use this instead:

```
no_browser_config = Config.no_browser
Config.no_browser = true
...
Config.no_browser = no_browser_config
```

lib/rubycritic/commands/compare.rb
```diff
@@ -0,0 +1,115 @@
...
+      def threshold_reached?
+        (Config.base_branch_score - Config.feature_branch_score) > Config.threshold_score
+      end
+
+      def branch_directory(branch)
+        "tmp/rubycritic/compare/#{Config.send(branch)}"
+      end
+
+      def build_directory
+        "tmp/rubycritic/compare/builds/build_#{@build_number}"
```
Onumis: This also does not respect the `--path` argument. Should be `"#{Config.root}/compare/builds/build_#{@build_number}"`

lib/rubycritic/commands/compare.rb
```diff
@@ -0,0 +1,115 @@
...
+      def mark_build_fail?
+        Config.threshold_score > 0 && threshold_reached?
+      end
+
+      def threshold_reached?
+        (Config.base_branch_score - Config.feature_branch_score) > Config.threshold_score
+      end
+
+      def branch_directory(branch)
+        "tmp/rubycritic/compare/#{Config.send(branch)}"
```
Onumis: This also does not respect the `--path` argument. Should be `"#{Config.root}/compare/#{Config.send(branch)}"`

lib/rubycritic/commands/compare.rb
```diff
@@ -0,0 +1,115 @@
...
+        analyse_branch(:feature_branch)
+        Config.no_browser = false
+        analyse_modified_files
+        compare_code_quality
+      end
+
+      # keep track of the number of builds and
+      # use this build number to create seperate directory for each build
+      def update_build_number
+        build_file_location = '/tmp/build_count.txt'
```
Onumis: This is creating a file under `/tmp`. This folder only exists on unix base systems.
Should be `"#{Config.root}/build_count.txt"`

lib/rubycritic/commands/compare.rb
```diff
@@ -0,0 +1,115 @@
...
+        Config.feature_root_directory = Pathname.new(branch_directory(:feature_branch))
+        Config.build_root_directory = Pathname.new(build_directory)
+      end
+
+      # switch branch and analyse files
+      def analyse_branch(branch)
+        SourceControlSystem::Git.switch_branch(Config.send(branch))
+        critic = critique(branch)
+        Config.send(:"#{branch}_score=", critic.score)
+        Config.root = branch_directory(branch)
```
Onumis: I don't think we should be overriding the arguments. At least make a copy and revert it after this run.

In fact, I'm thinking of removing the attr_writer methods for all the options. :thinking: 
If we get a bug related to the options, it'll be hell to track it down because we're changing them mid analysis.

lib/rubycritic/commands/compare.rb
```diff
@@ -0,0 +1,116 @@
...
+        Config.no_browser = true
+        analyse_branch(:base_branch)
+        analyse_branch(:feature_branch)
+        Config.no_browser = original_no_browser_config
+        analyse_modified_files
+        compare_code_quality
+      end
+
+      # keep track of the number of builds and
+      # use this build number to create seperate directory for each build
```
Onumis: typo: `separate`

lib/rubycritic/commands/compare.rb
```diff
@@ -0,0 +1,116 @@
...
+      def update_build_number
+        build_file_location = "#{Config.root}/build_count.txt"
+        File.new(build_file_location, 'a') unless File.exist?(build_file_location)
```
Onumis: This is causing the tests to fail. We need to ensure the directory exists:

`FileUtils.mkdir_p(Config.root) unless File.directory?(Config.root)`

Onumis:
Manually fixed other issues and pushed to master.

PR #194, opened 2016-11-29
Add missing method 'head_reference' for Perforce source control

Rataah:
Hi guys,

This PR is related to this [pull request](https://github.com/whitesmith/rubycritic/pull/165).
I missed a method named "head_reference" in Perforce source control class. 

Since there is no "head commit" in a Perforce workspace, the reference given is the latest "changelist" number of the directories covered by RubyCritic.

Even if this method is enable now, I don't see any differences in the generated HTML files.

Cheers


Onumis:
Again, not much I can do to manually test this, but it looks good! :D

PR #196, opened 2016-12-02
Add MRI 2.3.3 compatibility

josephpage:
I just updated the `parser` gem to [2.3.3.1](https://github.com/whitequark/parser/blob/master/CHANGELOG.md)


tejasbubane:
Shouldn't we also make sure travis is good with `2.3.3`?

Onumis:
Yep, I'll update travis.yml to ensure that.

tejasbubane:
@Onumis Done #199

PR #198, opened 2016-12-03
Running "rake rubycritic" in rails project does not display gpa-chart(in RubyCritic version 3.1.0)

hoshinotsuyoshi:
Hi guys,

New UI/UX of rubycritic(version 3.1.0) is awesome!! :tada: thx!

When I was going to upgrade "rubycritic" to version 3.1.0 in my rails project, I found that running `rake rubycritic` did not display gpa-chart.

I think that there is an interference problem between ActiveSupport's `Object#to_json` extension and `RubyCritic::Rating#to_json`.

### Reproducing

Versions:

* rails 5.0.0
* rubycritic 3.1.0

I use `lib/tasks/rubycritic.rake` like below:

```
require 'rubycritic/rake_task'
RubyCritic::RakeTask.new
```

I run:

```
$ bin/rake rubycritic
```

Results in:

**bad**
bad

There are 2 bad points:

* Not displaying gpa-chart
* Colorless plots

But of course

```
$ bundle exec rubycritic .
```

will work well :). This outputs good html.

**good**
good

### overview.html source difference

Running `bin/rake rubycritic`(bad) v.s. Running `bundle exec rubycritic .`(good)

```diff
...(snip)....
 <script type="text/javascript">
   var turbulenceData = [
-    {"name":"ApplicationController","x":3,"y":0,"rating":{"letter":"A"}},
-    {"name":"BucketsController","x":12,"y":7,"rating":{"letter":"A"}}
+    {"name":"ApplicationController","x":3,"y":0,"rating":"A"},
+    {"name":"BucketsController","x":12,"y":7,"rating":"A"}
     ...(snip)....
   ];
   var score = 98.95833333333333;
 </script>
...(snip)....
```

So it seems that rendering `{"letter":"A"}` is bad.

### To fix it

Enforcing `rating` value to be a string value in `RubyCritic::Turbulence.data` solves this problem.

### DIFF
--- a/lib/rubycritic/generators/html/turbulence.rb
+++ b/lib/rubycritic/generators/html/turbulence.rb
@@ -8,7 +8,7 @@ module RubyCritic
           name: analysed_module.name,
           x: analysed_module.churn,
           y: analysed_module.complexity,
-          rating: analysed_module.rating
+          rating: analysed_module.rating.to_s
         }
       end.to_json
     end


Onumis:
@hoshinotsuyoshi Thanks for the bugfix and for the awesomely detailed explanation!

This LGTM, but since @tejasbubane and @rohitcy recently revamped the UI and changed this, they may have a different approach.

tejasbubane:
@rohitcy was able to reproduce this. What we were not able to figure was where was this `{"letter":"A"}` coming from? [`rating` has a `to_json` method](https://github.com/whitesmith/rubycritic/blob/master/lib/rubycritic/core/rating.rb#L26) which should return a string. And even Activesupport's `to_json` should call internally call this method itself.

Onumis:
@tejasbubane the thing is that here we are not calling `Rating#to_json` anywhere.

```
(byebug) analysed_module.rating
#<RubyCritic::Rating:0x0000000871f3d8 @letter="A">

(byebug) analysed_module.rating.to_s
"A"
(byebug) analysed_module.rating.to_json
"\"A\""

# what is currently happening
(byebug) [ {rating: analysed_module.rating} ].to_json
"[{\"rating\":{\"letter\":\"A\"}}]"

# fix proposed by @hoshinotsuyoshi
(byebug) [ {rating: analysed_module.rating.to_s} ].to_json
"[{\"rating\":\"A\"}]"
```

Although that a better fix might be to change the `AnalysedModule#to_h` method, to call `rating.to_s`:

```
    def to_h
      {
        name: name,
        path: path,
        smells: smells,
        churn: churn,
        committed_at: committed_at,
        complexity: complexity,
        duplication: duplication,
        methods_count: methods_count,
        cost: cost,
        rating: rating.to_s
      }
    end
```

Onumis:
Merging this in now, since it's the correct fix. thanks @hoshinotsuyoshi

hoshinotsuyoshi:
thank you!! ✨

PR #200, opened 2016-12-05
Drop the minor version from travis

Onumis:
Follow up from #199.

This way Travis will always test with the latest minor version, but sadly, `rvm: 2.3` is not working :\

@tejasbubane


tejasbubane:
> but sadly, rvm: 2.3 is not working

Yeah, this seems to be an issue with travis' build environment https://github.com/travis-ci/travis-ci/issues/5361#issuecomment-167536760

PR #201, opened 2016-12-08
Use a real progress bar for ruby critic

thedrow:
This PR introduces a real progress bar which greatly improves the UX of the command line.
The PR is not 100% complete but it is good enough to merge and release.

We need to centralize our reporting through one component so that the progress bar won't break.
I'll do that another time.

### DIFF
--- a/lib/rubycritic/analysers/smells/flay.rb
+++ b/lib/rubycritic/analysers/smells/flay.rb
@@ -13,6 +13,7 @@ module RubyCritic
 
       def run
         hashes = @flay.hashes
+        skip_logging = @logger.nil?
         hashes.each do |structural_hash, nodes|
           smell = create_smell(structural_hash, nodes)
           nodes.map(&:file).uniq.each do |file|
@@ -23,10 +24,10 @@ module RubyCritic
             @analysed_modules[node.file].duplication += node.mass
           end
 
-          @logger.report_completion unless @logger.nil?
+          @logger.report_completion unless skip_logging
         end
 
-        @logger.report_completion @analysed_modules.size - hashes.size unless @logger.nil?
+        @logger.report_completion @analysed_modules.size - hashes.size unless skip_logging
       end
 
       def to_s
--- a/lib/rubycritic/analysers/smells/flay.rb
+++ b/lib/rubycritic/analysers/smells/flay.rb
@@ -1,14 +1,13 @@
 # frozen_string_literal: true
 require 'rubycritic/analysers/helpers/flay'
 require 'rubycritic/core/smell'
-require 'rubycritic/colorize'
 
 module RubyCritic
   module Analyser
     class FlaySmells
-      include Colorize
-      def initialize(analysed_modules)
+      def initialize(analysed_modules, logger=nil)
         @analysed_modules = paths_to_analysed_modules(analysed_modules)
+        @logger = logger
         @flay = Flay.new(@analysed_modules.keys)
       end
 
@@ -22,9 +21,11 @@ module RubyCritic
           nodes.each do |node|
             @analysed_modules[node.file].duplication += node.mass
           end
-          print green '.'
+
+          @logger.report_completion unless @logger.nil?
         end
-        puts ''
+
+        @logger.report_completion @analysed_modules.size - @flay.hashes.size unless @logger.nil?
       end
 
       def to_s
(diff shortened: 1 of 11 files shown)


thedrow:
If you prefer I can create a singleton EmptyLogger that will do nothing instead of the NilCheck.

thedrow:
@Onumis care to take a look?

Onumis:
@thedrow this looks really cool! :D

I'd still like to see the result of each analyser.
Previously we had something like this:

```
running flay smells
..........................................................
running flog smells
..........................................................
running reek smells
..........................................................
running complexity
..........................................................
running attributes
..........................................................
running churn
..........................................................
```

With this PR we have:

```
Running churn: |============= 100 ============[354/354]>>| Time: 00:00:02 Time: 00:00:02
```

It'd be nice to leave logged in the console the progress of flay, flog, reek, etc.
Maybe as simple as:

```
Flay: [50/50]
Flog: [50/50]
Reek: [50/50]
...
Running churn: |============= 100 ============[354/354]>>| Time: 00:00:02 Time: 00:00:02
```

What do you think?

Also, there are a few code smells. After that and we can merge this.

thedrow:
I think I can do that. As for the code smell. I'm not sure how to resolve it.
Do we need an EmptyLogger as a default?

troessner:
>> As for the code smell. I'm not sure how to resolve it.

You can implement a NullLogger following the [NullObject pattern](https://robots.thoughtbot.com/rails-refactoring-example-introduce-null-object) and then use dependency injection in the constructor like this:

```
def initialize(analysed_modules, logger = NullLogger)
end
```

Other than that: I believe we should remove some of the code duplication before merging this one but I'll rather comment in the PR.

lib/rubycritic/analysers/smells/flay.rb
```diff
@@ -1,19 +1,20 @@
...
-      def initialize(analysed_modules)
+      def initialize(analysed_modules, logger=nil)
         @analysed_modules = paths_to_analysed_modules(analysed_modules)
+        @logger = logger
         @flay = Flay.new(@analysed_modules.keys)
       end
 
       def run
-        @flay.hashes.each do |structural_hash, nodes|
+        hashes = @flay.hashes
```
troessner: You're using `hashes` just one time here so it doesn't add much value. I'd rather drop it and use `@flay.hashes` directly below.

lib/rubycritic/analysers/smells/flay.rb
```diff
@@ -1,19 +1,20 @@
...
+        skip_logging = @logger.nil?
```
troessner: That and everything associated to it below can go away with a NullLogger (see my other comment).

lib/rubycritic/analysers_runner.rb
```diff
@@ -23,11 +24,16 @@ def initialize(paths)
...
-        analyser_instance = analyser_class.new(analysed_modules)
-        puts "running #{analyser_instance}"
+        analyser_instance = analyser_class.new(analysed_modules, logger)
+        logger.topic = analyser_instance
+
         analyser_instance.run
       end
+
+      puts ''
+
```
troessner: Let's drop the newline here

lib/rubycritic/core/analysed_modules_collection.rb
```diff
@@ -30,6 +30,10 @@ def to_json(*options)
       @modules.to_json(*options)
     end
 
+    def size
+      @modules.size
```
troessner: I'd prefer using delegation via Ruby's Forwardable module here to communicate our intent clearer.

lib/rubycritic/analysers/smells/reek.rb
```diff
@@ -1,22 +1,21 @@
...
-require 'rubycritic/colorize'
 
 module RubyCritic
   module Analyser
     class ReekSmells
-      include Colorize
-      def initialize(analysed_modules)
+      def initialize(analysed_modules, logger=nil)
         @analysed_modules = analysed_modules
+        @logger = logger
```
troessner: Now that is my biggest problem with this PR. This line and the constructor line and the line below are repeated in every analyzer. So like 5 or 6 times which is a pretty big smell in my mind. 

We either should use a new module named "Loggable" here or - and that would be my preferred choice - create a "BaseAnalyzer" from which all analyzers inherit. This would allow us to remove most, if not all the duplication.

Onumis:
@thedrow do you think we can get this finished? :smile:

thedrow:
I no longer use Ruby so I lost interest in this PR.
Feel free to complete it. There's not much to do.

Onumis:
closing due to inactivity

PR #203, opened 2016-12-15
Fix errors when no source-control is found

### LINKED ISSUE #188: Error occurs when no source control is used
Hi,

I tried the new version on one of my project with no source control enabled and an error occurs:

```
O:/ruby/ruby-2.3.1-x64-mingw32/lib/ruby/gems/2.3.0/gems/rubycritic-3.1.0/lib/rubycritic/analysis_summary.rb:33:in `+': String can't be coerced into Fixnum (TypeError)
        from O:/ruby/ruby-2.3.1-x64-mingw32/lib/ruby/gems/2.3.0/gems/rubycritic-3.1.0/lib/rubycritic/analysis_summary.rb:33:in `block in churns'
        from O:/ruby/ruby-2.3.1-x64-mingw32/lib/ruby/gems/2.3.0/gems/rubycritic-3.1.0/lib/rubycritic/analysis_summary.rb:33:in `each'
```

The error is raised in the following method of AnalysisSummary class:
```ruby
    def churns
      modules.inject(0) { |acc, elem|
        acc + elem.churn 
      }
    end
```

Where the value of "**element.churn**" is the string "**N/A"**
The "**N/A**" value seems to come from the Double class in SourceControlSystem.
Precisely:
```ruby
      def revisions_count(_)
        'N/A'
      end
```

Ruby version: 
```
> ruby -v
ruby 2.3.1p112 (2016-04-26 revision 54768) [x64-mingw32]

```
Rubycritic version:
```
> gem list rubycritic

*** LOCAL GEMS ***

rubycritic (3.1.0)
```

Many thanks for this gem and keep up your excellent work!

tejasbubane:
Fixes #188 

Churn calculation requires source-control.
Do not show the churn-complexity graph when source control is not found.

screen shot 2016-12-16 at 1 12 33 am

### DIFF
--- a/lib/rubycritic/generators/html/templates/overview.html.erb
+++ b/lib/rubycritic/generators/html/templates/overview.html.erb
@@ -20,12 +20,16 @@
         <!--Churn vs Complexity graph-->
         <div class="col-xs-7 col-sm-7 col-md-6 col-lg-8">
           <div class="Graph_Cards fadeIn">
+            <% if Config.source_control_present? %>
+              <div id="churn-vs-complexity-graph-container" class="chart-container"></div>
+            <% else %>
+              <div id="churn-error">We can't show you Churn-vs-Complexity graph because this project does not seem to be under a source code management system like git or perforce at the moment. This doesn't mean that anything is wrong, it just means certain features like this one are not avaiable.</div>
+            <% end %>
 
-          <div id="churn-vs-complexity-graph-container" class="chart-container"></div>
-          <script type="text/javascript">
-            var turbulenceData = <%= @turbulence_data %>;
-            var score = <%= @score %>;
-          </script>
+            <script type="text/javascript">
+             var turbulenceData = <%= @turbulence_data %>;
+             var score = <%= @score %>;
+            </script>
           </div>
         </div>
         <!--End Churn vs Complexity graph-->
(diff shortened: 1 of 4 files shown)


Onumis:
LGTM!

lib/rubycritic/generators/html/templates/overview.html.erb
```diff
@@ -20,12 +20,16 @@
         <!--Churn vs Complexity graph-->
         <div class="col-xs-7 col-sm-7 col-md-6 col-lg-8">
           <div class="Graph_Cards fadeIn">
+            <% if Config.source_control_present? %>
+              <div id="churn-vs-complexity-graph-container" class="chart-container"></div>
+            <% else %>
+              <div id="churn-error">Please use a source-control system to calculate Churn.</div>
```
troessner: That's a very misleading error message. I don't think it should be an error at all since the user did nothing wrong.

How about changing this to

>> We can't show you churn because this project does not seem under a source code management system like git or perforce at the moment. This doesn't mean that anything is wrong, it just means certain features like this one are not avaiable.

troessner:
1 small thing and then LGTM. And one of these days we might even have tests for all of this ;)

tejasbubane:
@troessner Fixed. I should learn to write better docs 😄

troessner:
Very nice, merged, thanks!

PR #217, opened 2017-02-24
Update RuboCop to 0.47.1 from 0.42.0.

jdickey:
Obvious change to `rubycritic.gemspec`. This commit also changes the
`.rubocop.yml` file to disable a `Security/MarshalLoad` cop (introduced
in RuboCop 0.47.0) from flagging the `RubyCritic::Serializer.load`
method. Interestingly, wrapping that method in RuboCop disable/enable
comments for that cop simply flags a `Lint/UnneededDisable` violation.
Go figure. Adding the one file to a disable-Security/MarshalLoad entry
in `.rubocop.yml` solves the problem, even if it does feel like shooting
a fly with an RPG.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -18,4 +18,9 @@ Metrics/LineLength:
 Style/Documentation:
   Enabled: false
 
+Security/MarshalLoad:
+  Enabled: false
+  Include:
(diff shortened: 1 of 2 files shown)


Onumis:
Thank you!

PR #220, opened 2017-02-25
Add YARD options file, render images in generated docs

olleolleolle:
This PR adds a configuration file for YARD, the API docs generator.

To make the documentation pretty:

  - README: image tags with relative links, so that it can render elsewhere, too.
  - Generated docs: copy `images/` directory as "assets" to the output folder (`doc/`)

### DIFF
--- a/.yardopts
+++ b/.yardopts
@@ -1,0 +1,5 @@
+--no-private
+--use-cache
+--asset images
+lib/**/*.rb -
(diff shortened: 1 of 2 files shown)


Onumis:
Thank you!

Onumis:
oh, I though it'd look good here also: http://www.rubydoc.info/github/whitesmith/rubycritic

I guess we need to use absolute paths...

PR #222, opened 2017-03-10
Update travis to specifically test against ruby 2.4

### LINKED ISSUE #221: Ruby 2.5.0dev is out - update travis to specifically test agains 2.4
Context: https://github.com/whitesmith/rubycritic/pull/219#issuecomment-283403287

Onumis:
Close #221

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,14 @@ matrix:
     - rvm: 2.1
     - rvm: 2.2
     - rvm: 2.3.3
+    - rvm: 2.4.0
     - rvm: ruby-head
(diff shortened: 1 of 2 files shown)


Onumis:
@olleolleolle done!

tejasbubane:
@Onumis Can you add `2.4` to the compatibility section in README?

Also we have only one failing test for jruby. We should open another issue to fix that. Then we can have support for JRuby as well 🎉  😄

olleolleolle:
The smelly.rb which has a `method_missing` that uses `define_method` - try excluding that once. It _could_ be the JRuby culprit.

PR #228, opened 2017-03-20
Tweak CHANGELOG.md since v3.1.3 doesn't support Ruby 2.4.0

yuku-t:
It seems that master supports Ruby 2.4.0 (thanks for @jdickey 👏 👏 ) but v3.1.3 doesn't.

And, you know, three months has already passed since Ruby 2.4.0 was out. Could you release the next version?


Onumis:
Thanks you! I'll release a new version today.

PR #231, opened 2017-04-14
Update the "Getting Started"

koic:
RubyCritic supports Ruby 2.1 or higher. I think that Ruby 1.9's hash syntax is good for sample on how to write to Gemfile.


koic:
This build error aren't related to above change. I think that will be solved by PR #232.

PR #232, opened 2017-04-22
Fix a RuboCop offenses

koic:
This PR fixes the following offenses.

```sh
% rubocop
Inspecting 99 files
CCC.CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC.CCCCCCC..CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC

Offenses:

(snip)

99 files inspected, 107 offenses detected
```

cf. https://travis-ci.org/whitesmith/rubycritic/jobs/221954186

This is what I did with PR. First, I fixed it using autocorrect (`rubocop -a`) . Next, it disable Style/IndentHeredoc cop. It is disabling Style/IndentHeredoc cop for the following reasons.

- RubyCritic does not depend on Active Support, then it can not use `String#strip_heredoc`.
- RubyCritic has support of Ruby 2.2 or lower, then it can not use squiggly heredoc (`<<~`) .

Thank you.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -40,3 +40,6 @@ Style/MethodMissing:
   Exclude: 
     - 'lib/rubycritic/configuration.rb'
 
+# Offense count: 3
+Style/IndentHeredoc:
+  Enabled: false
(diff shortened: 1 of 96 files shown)


Onumis:
Many thanks!

PR #237, opened 2017-05-13
Only scroll Y axis when visiting page with anchor tag

georgedrummond:
### Problem

`$.scrollTo` was using [default setting](https://github.com/flesler/jquery.scrollTo#settings) of scrolling to `xy` axis to scroll to line number. On my 13 inch screen this would cause a weird rendering issue where the page score would be partially overlapped. 

### Solution

Only `$.scrollTo` in the `y` axis.

### Example

ruby_critic_-_home


Onumis:
Thank you!

PR #238, opened 2017-05-18
Relax dependency version of parser gem

yuku-t:
My Rails project uses Ruby 2.4.1 so that parser v2.4.0 warns as follows:

image

This patch fixes it.


Onumis:
Sounds good, but we'll have to wait for: https://github.com/whitequark/parser/issues/346 :)

Issue #230, opened 2017-03-27
Highcharts turboThreshold not configurable


AzizHarrington:
The current value (1500) is too small for a large monolith, and prevents complexity/churn scatterplot from rendering. This should be configurable or set to a very large number.

Onumis:
Fixed by disabling the `turboThreshold`.

PR #241, opened 2017-05-19
Update application.js

ochagata:
Solves #230

### DIFF
--- a/lib/rubycritic/generators/html/assets/javascripts/application.js
+++ b/lib/rubycritic/generators/html/assets/javascripts/application.js
@@ -49,7 +49,7 @@ $("#churn-vs-complexity-graph-container").highcharts({
   },
   plotOptions: {
     series: {
-      turboThreshold: 1500
+      turboThreshold: 0
     },
     scatter: {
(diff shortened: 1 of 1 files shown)


lib/rubycritic/generators/html/assets/javascripts/application.js
```diff
@@ -49,7 +49,7 @@ $("#churn-vs-complexity-graph-container").highcharts({
   },
   plotOptions: {
     series: {
-      turboThreshold: 1500
+      turboThreshold: 15000
```
Onumis: I'd rather set this to `0` so this won't ever be a problem.

I'm also thinking on how we could make this configurable :thinking: 
Maybe add a control on the page to set this value and re-render the chart. But not sure if it'd be that useful. 

@AzizHarrington what do you think?

ochagata:
@Onumis Made the change

ochagata:
@Onumis mind merging this, please?

Onumis:
will make a release later

ochagata:
@Onumis Awesome

PR #242, opened 2017-05-20
Traverse git history only once

joneshf:
# Why this PR?

This is specific to my problem, but it probably applies to repos of other shapes and sizes along with different version control systems.

We've got a five year old repo (to the day 🎂), with 46k commits over 9k files. I mention these numbers only to provide context. When churn is run over the whole repo, it takes about 30 minutes to complete. I think this is because it traverses the entire history for each file path given. Unfortunately if that's the performance of churn, we can't start using RubyCritic in CI.

I've started experimenting with traversing the history once, caching the results and using the cache for quicker lookup. Using this method, churn takes about 2 minutes to complete. That's something we can deal with on CI.

# What's the change?

Before, we were traversing the history for each path.
That behavior gets costly with a long history.

Now, we are traversing the history once initially,
creating a cache from paths to revision counts,
and referencing the cache for each path.

# Further ideas

We might be able to do similar for the other source control systems.
However, I personally don't have enough experience with them,
nor do I have an actual project to test on.

I welcome any critiques or other ideas.

### DIFF
--- a/lib/rubycritic/source_control_systems/git.rb
+++ b/lib/rubycritic/source_control_systems/git.rb
@@ -14,7 +14,7 @@ module RubyCritic
       end
 
       def revisions_count(path)
-        `git log --follow --format=%h #{path.shellescape}`.count("\n")
+        path_revisions_count.fetch(path.shellescape, 0)
       end
 
       def date_of_last_commit(path)
@@ -51,6 +51,18 @@ module RubyCritic
 
       def travel_to_original_state
         `git stash pop`
+      end
+
+      def path_revisions_counts
+        `git log --name-only --format='' | sort | uniq -c`
+      end
+
+      def path_revisions_count
+        @path_revisions_count ||=
+          path_revisions_counts.split("\n").map do |string|
+            count, path = string.split
+            [path, count.to_i]
+          end.to_h
       end
     end
   end
--- a/lib/rubycritic/source_control_systems/git.rb
+++ b/lib/rubycritic/source_control_systems/git.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'rubycritic/source_control_systems/git/revision'
+
 module RubyCritic
   module SourceControlSystem
     class Git < Base
@@ -53,16 +55,21 @@ module RubyCritic
         `git stash pop`
       end
 
-      def path_revisions_counts
-        `git log --name-only --format='' | sort | uniq -c`
+      def name_status
+        `git log --all --name-status --format='' --reverse`
       end
 
       def path_revisions_count
-        @path_revisions_count ||=
-          path_revisions_counts.split("\n").map do |string|
-            count, path = string.split
-            [path, count.to_i]
-          end.to_h
+        @path_revisions_count ||= path_revisions_cache
+      end
+
+      def path_revisions_cache
+        revisions = [Add, Copy, Delete, Modify, Rename, Double]
+
+        name_status
+          .split("\n")
+          .map { |string| Revision.parse(revisions, string) }
+          .each_with_object({}) { |revision, hash| revision.evaluate(hash) }
       end
     end
   end
--- a/lib/rubycritic/source_control_systems/git/revision.rb
+++ b/lib/rubycritic/source_control_systems/git/revision.rb
@@ -1,0 +1,88 @@
+# frozen_string_literal: true
+
+module RubyCritic
+  module SourceControlSystem
+    class Git < Base
+      class Revision
+        def initialize(path, new_path)
+          @path = path
+          @new_path = new_path
+        end
+
+        def self.matches?(status)
+          match =~ status
+        end
+
+        def self.parse(revisions, string)
+          status, path, new_path = string.split("\t").map(&:shellescape)
+          revisions
+            .find { |revision| revision.matches?(status) }
+            .new(path, new_path)
+        end
+
+        private
+
+        attr_reader :path, :new_path
+      end
+
+      class Double < Revision
+        def evaluate(_) end
+
+        def self.match
+          //
+        end
+      end
+
+      class Add < Revision
+        def evaluate(hash)
+          hash[path] = 1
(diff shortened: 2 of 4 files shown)


lib/rubycritic/source_control_systems/git.rb
```diff
@@ -52,6 +52,18 @@ def stashes_count
       def travel_to_original_state
         `git stash pop`
       end
+
+      def path_revisions_counts
+        `git log --name-only --format='' | sort | uniq -c`
```
guilhermesimoes: 1. `uniq` is exclusive to unix machines so this won't run on windows.
2. The [option `follow`](https://www.git-scm.com/docs/git-log#git-log---follow) was used to provide a more accurate history of a file given that it tracks changes beyond renames. Without this option the `churn` score won't be as accurate as it used to be. Unfortunately, this option only works for a single file, so we can't use it here. This may be a trade-off we're willing to make or not...

joneshf: Thanks for the heads up!

1. Seems fixable if we process the output of the `git` command in ruby, rather than using `sort` and `uniq`. But I think the second issue will negate this one.
1. Ah, I didn't think about that! Seems possible to get around as well—though slightly more work. We can grab the name of the file along with what kind of change it was:
    ```shell
    $ git log --all --format='' --name-status --reverse
    ...
    A rubycritic.gemspec
    M rubycritic.gemspec
    M Rakefile
    M lib/rubycritic/report_generators/file_generator.rb
    M lib/rubycritic/report_generators/line_generator.rb
    R081  lib/rubycritic/report_generators/report_generator.rb  lib/rubycritic/report_generators/reporter.rb
    M lib/rubycritic/report_generators/base_generator.rb
    A lib/rubycritic/report_generators/index_generator.rb
    M lib/rubycritic/report_generators/reporter.rb
    ...
    ```
    Then parse the lines and build up the cache of changes:
    * A - add the path to the hash with value 1
    * B - ignore
    * C - add the copied path to the hash with value 1
    * D - remove the path from the hash
    * M - increment path value by 1
    * R - rename the path in the hash and increment the value by 1
    * T - increment path value by 1
    * U - ignore
    * X - ignore

    Since we log in reverse order, it should handle the case where a file named `foo` was renamed to `bar`, and later a new `foo` is introduced.

    Does that make sense?

guilhermesimoes: Interesting, I think it makes sense!

I just tried the `git log` command you suggested in a big repository and got these two warnings:

> warning: inexact rename detection was skipped due to too many files.
> warning: you may want to set your diff.renameLimit variable to at least 2098 and retry the command.

Not sure if we want to:

1. Suppress these warnings.
2. Fix the underlying cause of these warnings (I found [this](https://github.com/grimoirelab/perceval/issues/19) but not sure if it's an optimal solution).
3. Do nothing and leave it like this.

Don't let this discourage you though. If it's not a big effort try your suggested approach. I'd appreciate it if you could then compare your previous `churn` scores with the new scores.

Onumis: I'd like to add that we should scope this to the paths passed to rubycritic.
For example, if I do `rubycritic ./app` we shouldn't care about the `git log` of files outside of this dir, right?
(or maybe we do because of possible file moves?)

joneshf: @guilhermesimoes Interesting! I'll try and check to see what that does on a repo that's under the limit. It's not clear to me if that would also cause a slowdown on a small repo.

@Onumis I think you hit the nail on the head. Someone could have renamed a file from a directory they ignore now into a directory they care about. Might not be that common of an occurrence though. I think it might be best to play it safe and look at everything. What do you think?

Onumis: well, this is the kind of tool that should favour correctness over speed, so let's look at everything.

joneshf:
Sorry for letting this flounder.

Pushed a change that implements the algorithm we talked about. Mind taking a quick glance over and seeing if I'm spinning off into left field? I tried to follow the patterns laid out in other places of the codebase.

Speed seems about the same as the previous commit. Data looks like it's pretty much the same. Some files have slightly more commits than before, some have slightly less. I'm not sure if that's because of the way `follow-name` works or if there's a problem with what I've implemented. I'm going to guess the latter 😉.

I'll add some tests for this stuff and fix the build if the direction I'm going isn't too far off.

Onumis:
LGTM!

joneshf:
Hmm. I've tried running `bundle exec rake` like travis does on `ruby-2.1.0` and `ruby-2.3.3`. It looks to pass all the tests. Going to close and re-open to force a build on travis.

joneshf:
Didn't seem to change anything. I'm not sure how to proceed with the failing tests, but will look at it in a bit.

lib/rubycritic/source_control_systems/git/revision.rb
```diff
@@ -0,0 +1,88 @@
...
+
+        def self.matches?(status)
+          match =~ status
+        end
+
+        def self.parse(revisions, string)
+          status, path, new_path = string.split("\t").map(&:shellescape)
+          revisions
+            .find { |revision| revision.matches?(status) }
+            .new(path, new_path)
```
Onumis: @joneshf I think the test errors come from here. I think that in some strange case, the `.matches?` never returns true.

Maybe status is nil, or empty?

Onumis: @joneshf bump! hacktoberfest is just around the corner :wink:

joneshf: Sorry for leaving this open for like four months 😅.

Work is letting up a bit so I can probably jump back on it soon. If I don't get to it within the next two weeks, I'll close it since I probably won't be finishing it up.

joneshf:
I'm going to revisit this after the new year. Sorry for letting this sit for months.

PR #247, opened 2017-10-05
Make json_report_test test on actual files

nightscape:
We're currently creating a new reporter and along the way noticed that `json_report_test.rb` does not test anything semantically.
This test makes `json_report_test.rb` an integration test that actually runs analyzers on the FakeFS and checks that all sample files appear in the report.

### DIFF
--- a/lib/rubycritic/source_control_systems/git.rb
+++ b/lib/rubycritic/source_control_systems/git.rb
@@ -1,12 +1,23 @@
 # frozen_string_literal: true
 
+require 'tty/which'
+
 module RubyCritic
   module SourceControlSystem
     class Git < Base
       register_system
+      GIT_EXECUTABLE = TTY::Which.which('git')
+
+      def self.git(arg)
+        `#{GIT_EXECUTABLE} #{arg}`
+      end
+
+      def git(arg)
+        self.class.git(arg)
+      end
 
       def self.supported?
-        `git branch 2>&1` && $?.success?
+        git('branch 2>&1') && $?.success?
       end
 
       def self.to_s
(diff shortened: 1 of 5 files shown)


lib/rubycritic/source_control_systems/git.rb
```diff
@@ -1,32 +1,43 @@
 # frozen_string_literal: true
 
+require 'tty/which'
+
 module RubyCritic
   module SourceControlSystem
     class Git < Base
       register_system
+      GIT_EXECUTABLE = TTY::Which.which('git')
```
Onumis: Nice!

test/test_helper.rb
```diff
@@ -18,6 +20,19 @@ def capture_output_streams
...
+    begin
+      config = File.expand_path('..', __dir__)
+      FakeFS::FileSystem.clone(config)
+      Dir.chdir(config)
+      yield
+    ensure
+      FakeFS::FileSystem.clear
+    end
+  end
+end
```
Onumis: :ok_hand:

lib/rubycritic/source_locator.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'pathname'
+require 'rubycritic/configuration'
```
Onumis: Why this require?

Onumis:
Looks great! Just one small question, but overall this is mergable already

lib/rubycritic/source_locator.rb
nightscape: If I remember correctly, the tests failed without this when I ran them outside of `rake` like
```
bundle exec ruby -Ilib:test test/lib/rubycritic/generators/json_report_test.rb 
```
Personally, I also consider it good style to have Ruby files be "self-sufficient", i.e. you can `require` them without having to `require` other files first, but that might be a matter of taste.

Onumis:
Great, thanks!

PR #248, opened 2017-10-05
Lint report

nightscape:
This PR is based on https://github.com/whitesmith/rubycritic/pull/247 and adds a reporter which writes files in a format that many other linters (like e.g. Golint) use.
We are using it in combination with https://github.com/jenkinsci/violation-comments-to-github-plugin to create Github PR comments with rubycritic's output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # master [(unreleased)](https://github.com/whitesmith/rubycritic/compare/v3.2.3...master)
+* [FEATURE] Add lint format similar to Golint (by [@nightscape][])
 
 # 3.2.3 / 2017-05-31 [(commits)](https://github.com/whitesmith/rubycritic/compare/v3.2.2...v3.2.3)
 
@@ -215,3 +216,4 @@
 [@georgedrummond]: https://github.com/georgedrummond
 [@yuku-t]: https://github.com/yuku-t
 [@ochagata]: https://github.com/ochagata
+[@nightscape]: https://github.com/nightscape
--- a/docs/jenkins-pr-reviews.md
+++ b/docs/jenkins-pr-reviews.md
@@ -1,0 +1,64 @@
+# Making Jenkins review Pull Requests
+
+## Installing Jenkins and setting up RubyCritic
+
+There is a step-by-step tutorial on how to set up Jenkins in [`building-own-code-climate.md`](./building-own-code-climate.md).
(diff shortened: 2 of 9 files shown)


Onumis:
Let's wait for #247 and then rebase this.

By the way, you'd receive many kudos if you were to... say, write a small guide on how to do that jenkins+GH+rubycritic thing, an place it on /docs :wink: :wink:

Onumis:
Don't forget to add an entry to `CHANGELOG.md` on the unreleased section (at the top)

nightscape:
Hi @Onumis,

I rebased on latest master and added an entry to the `CHANGELOG.md`.
Regarding the `docs/`, shall I put it into `building-own-code-climate.md`?

Onumis:
@nightscape I'd say it can be a separate file. Can also be on another PR, unless you have it ready already.

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # master [(unreleased)](https://github.com/whitesmith/rubycritic/compare/v3.2.3...master)
+* [FEATURE] Add lint format similar to Golint
```
Onumis: please add your handle here: `[@nightscape][]`

and at the bottom of the file: `[@nightscape]: https://github.com/nightscape`

nightscape:
I also thought about naming it `Linter` instead, but `Lint` seemed to match the current naming schema better (which mostly names what is being reported rather than who reports it).
But I also have no strong tendency either way. If you want I'll rename it.

CHANGELOG.md
nightscape: Done.

nightscape:
I just noticed that the used rule is not added properly. Give me another 5 minutes to fix that, then the PR should be good to go.

nightscape:
Ok, it's actually not that easy... There are not many `.txt` based lint formats that support a RULE and when they do they have an ugly syntax for it...

So from my side let's merge this as it is.

Maybe one day having a generic templating system where the user specifies an output template would be nice, but for our use-case it's overkill.

nightscape:
Hi @Onumis,

do you see any other open issues here, or can it be merged? :smile:

Onumis:
Nope, looks good! thanks. Will try to make a release later today

nightscape:
Great! Thanks for the rapid and useful feedback and the overall nice PR experience :smiley:

Issue #251, opened 2017-10-06
Prepare for ruby 2.5


Onumis:
[Our CI](https://travis-ci.org/whitesmith/rubycritic) is already complaining at some stuff that will break our build on ruby 2.5:

```

  1) Failure:
RubyCritic::Command::StatusReporter::with minimum-score option::when score is above the minimum#test_0001_should return the correct status [/home/travis/build/whitesmith/rubycritic/test/lib/rubycritic/commands/status_reporter_test.rb:77]:
Expected: "Score: 100.0"
  Actual: "Score: 100"

  2) Failure:
RubyCritic::Command::StatusReporter::with minimum-score option::when score is equal the minimum#test_0001_should return the correct status [/home/travis/build/whitesmith/rubycritic/test/lib/rubycritic/commands/status_reporter_test.rb:68]:
Expected: "Score: 99.0"
  Actual: "Score: 99"

  3) Failure:
RubyCritic::Command::StatusReporter::with default options#test_0002_accept a score [/home/travis/build/whitesmith/rubycritic/test/lib/rubycritic/commands/status_reporter_test.rb:26]:
Expected: "Score: 50.0"
  Actual: "Score: 50"

  4) Failure:
RubyCritic::Command::StatusReporter::with minimum-score option::when score is below minimum#test_0001_should return the correct status [/home/travis/build/whitesmith/rubycritic/test/lib/rubycritic/commands/status_reporter_test.rb:53]:
Expected: "Score (98.0) is below the minimum 99.0"
  Actual: "Score (98) is below the minimum 99.0"

```

tejasbubane:
This can be closed.

PR #252, opened 2017-10-08
Update tests for ruby 2.5

broder:
Should partially address #251 which looks to be caused by:

> Integer#{round,floor,ceil,truncate} now always return an Integer.

from https://docs.ruby-lang.org/en/trunk/NEWS.html.

I don't think rubocop supports Ruby 2.5 yet (https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/config.rb#L20) so there will still be more work to do on this issue.

### DIFF
--- a/lib/rubycritic/core/analysed_modules_collection.rb
+++ b/lib/rubycritic/core/analysed_modules_collection.rb
@@ -9,13 +9,13 @@ module RubyCritic
 
     # Limit used to prevent very bad modules to have excessive impact in the
     # overall result. See #limited_cost_for
-    COST_LIMIT = 32
+    COST_LIMIT = 32.0
     # Score goes from 0 (worst) to 100 (perfect)
-    MAX_SCORE = 100
+    MAX_SCORE = 100.0
     # Projects with an average cost of 16 (or above) will score 0, since 16
     # is where the worst possible rating (F) starts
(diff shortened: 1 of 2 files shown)


Onumis:
Whoa! That was a fast PR :stuck_out_tongue: 

Ruby 2.5 is still in an alpha-ish state, so it might take awhile. 

Anyway, it makes sense to explicitly change these cost-related vars to floating point.

Onumis:
@broder It's almost here! :tada: 
https://www.ruby-lang.org/en/news/2017/10/10/ruby-2-5-0-preview1-released/

As soon as this becomes available on rvm, we can update the travis build and run the tests again.
(https://github.com/rvm/rvm/pull/4204)

olleolleolle:
@Onumis Right, it _is_ now available on rvm. 

Q: Is 2.5. aliased to 2.5.0-preview1, now, is that what I'm seeing in `config/known` in that PR?

Onumis:
@olleolleolle yes, but bear in mind that it's not yet released: https://github.com/rvm/rvm/blob/master/CHANGELOG.md#unreleased

And even then, we'll have to wait a bit for Travis to update their RVM version :stuck_out_tongue: 
Patience!

nicolasleger:
@Onumis Ruby 2.5 is now released and available on Travis.

Issue #250, opened 2017-10-06
Solve Rubocop warnings


nunosilva800:
Go through `.rubocop_todo.yml` and fix those issues :) 

Just remove a line, run `rubocop` and you should see the warning/error.

zhikiri:
@Onumis I'll check it out

ericboehs:
I'll run the auto correct for `SpecialGlobalVars`.

zhikiri:
@ericboehs you can do everything, I'm really busy now

Onumis:
@tibbon if you are using latest rubocop, then we need to update the `AllCops/Include` setting in `.rubocop.yml` file

See https://github.com/bbatsov/rubocop/issues/4247#issuecomment-389364734

PR #254, opened 2017-10-13
RuboCop Auto Correct: Sytle/SpecialGlobalVars

ericboehs:


ericboehs:
It seems the auto correct for `$?` doesn't work for our scenario. Should I add instead add the TODO to our permanent ignore list?

Onumis:
@ericboehs Hi! Thanks for the contribution.
As a general note, you should reference the originating issue you are solving (in this case #250), and provide a brief description ;)

-----

You can make this work by adding a `require 'English'`.
You can place it in `lib/rubycritic/source_control_systems/base.rb`, right before `require 'shellwords'`.

Rubocop warns it when you run the autocorrect:
```
[Corrected] Prefer $CHILD_STATUS from the stdlib 'English' module (don't forget to require it) over $?.
```

See also: 
- https://ruby-doc.org/core-2.4.0/Process/Status.html
- http://ruby-doc.org/stdlib-2.4.0/libdoc/English/rdoc/English.html

Additionally, please add an entry to `CHANGELOG.md` on the unreleased section (at the top).

olleolleolle:
@ericboehs Hi! Do you still wish to make the suggested changes?

If yes, just commit more changes to this branch. If not, say the word, and we'll pick it up and carry it through.

Thanks again for making a contribution!

Onumis:
closing due to inactivity

PR #256, opened 2017-10-24
Clarify README

joshsaintjacque:
This is really minor, but I thought it might be good to clarify that the compatibility section of the README refers to Ruby.


Onumis:
Haha, true! Thanks :smile:

Issue #258, opened 2017-12-25
Timezone-dependent test-case: Perforce


olleolleolle:
There's a Perforce-related test which checks some dates: Perforce::Perforce::retrieve informations#test_0002_retrieves the date of the last commit of the ruby files

My environment's timezone is set to GMT+2.

The test fails like this:

```
  1) Failure:
Perforce::Perforce::retrieve informations#test_0002_retrieves the date of the last commit of the ruby files [/Users/olle/opensource/rubycritic/test/lib/rubycritic/source_control_systems/perforce_test.rb:156]:
Expected: "2016-09-05 11:39:11 +0000"
  Actual: "2016-09-05 13:39:11 +0200"
```

PR #259, opened 2017-12-25
perforce_test: Make test pass also in GMT+2

olleolleolle:
This PR makes a test pass also when the `TZ` isn't set to UTC.

See https://github.com/whitesmith/rubycritic/issues/258

It's my kludge workaround: make the `TZ=utc` while running the test.

---

Then, also: to make the build pass on Travis:

- `gem install bundler`, to have a working gem-based Bundler
- Pin RuboCop to 0.51.0, which has no current defects
- unroll rake tasks which want to `ruby bundle exec` themselves.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,5 +24,5 @@ matrix:
 script:
   - bundle exec rake test
   - bundle exec cucumber features --format progress --color
-  - bundle exec reek
+  - bundle exec rake reek
   - bundle exec rubocop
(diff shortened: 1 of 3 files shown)


olleolleolle:
Now, that worked in _2.1.0_. Why not the other ones/

olleolleolle:
@Onumis Would you want me to separate out the build-fixing things to their own PR?

Onumis:
@olleolleolle Nope, this is fine :)
Thanks!

PR #184, opened 2016-11-16
New Web UI

tejasbubane:
This is the implementation of the new UI being discussed here: https://github.com/whitesmith/rubycritic/issues/173
I also took the liberty to upgrade some of the javascript dependencies.

This is a combined effort from @cybrilla and we are loving it ❤️
Run it and let us know what you think.!

Closes: https://github.com/whitesmith/rubycritic/issues/173

PR #261, opened 2018-02-05
Fix wrong label for churn in the code file template

nbekirov:
This was changed during the New Web UI in PR #184.

Actually, I like the idea of "loc/method" metric but I'm unsure if and where to put it. For now I've just fixed the label.


Onumis:
Thanks!

PR #271, opened 2018-08-16
Branch comparison feature fix

antoineLeclercq:
Hi, this is a fix for the branch comparison feature. As of now, `-t`/`--maximum-decrease` does not behave as expected: it will not abort the process, aka exit the process with a code of 1, when the threshold is set to 0, even though the difference between the two branches' score is > 0.

I also tried to run rubocop before pushing and had to fix some issues in order to make it run. Please refer to commit message for more details.

### DIFF
--- a/lib/rubycritic/source_control_systems/git.rb
+++ b/lib/rubycritic/source_control_systems/git.rb
@@ -52,11 +52,11 @@ module RubyCritic
       end
 
       def self.switch_branch(branch)
-        uncommitted_changes? ? `git checkout #{branch}` : abort('Uncommitted changes are present.')
+        uncommitted_changes? ? abort('Uncommitted changes are present.') : `git checkout #{branch}`
       end
 
       def self.uncommitted_changes?
-        `git ls-files --other --exclude-standard --directory`.empty? && `git status --porcelain`.empty?
+        !`git ls-files --other --exclude-standard --directory`.empty? || !`git status --porcelain`.empty?
       end
 
       def self.modified_files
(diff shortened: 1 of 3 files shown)


lib/rubycritic/source_control_systems/git.rb
```diff
@@ -52,7 +52,13 @@ def self.switch_branch(branch)
...
 
       def self.uncommitted_changes?
-        `git ls-files --other --exclude-standard --directory`.empty? && `git status --porcelain`.empty?
+        git_ls_files = `git ls-files --other --exclude-standard --directory`
+        git_status = `git status --porcelain`
+        if !git_ls_files.empty? || !git_status.empty?
+          puts "There are uncommitted changes, FYI..."
+        end
+        # git_ls_files.empty? && git_status.empty?
+        true
```
Onumis: not sure I understand why this was changed. 
We can't just return true all the time, and the puts message is not needed since the caller `switch_branch` already aborts with a similar message.

We cannot do a `git checkout` when there are uncommitted changes.

antoineLeclercq: Oops, I think I forgot to branch out before pushing some debugging changes. Although, while debugging I found out that `uncommitted_changes?` returns the opposite of what it means. It should be `!git_ls_files.empty? || !git_status.empty?`, or the method should be renamed something like `no_uncommited_changes?`

antoineLeclercq:
@Onumis PTAL
I rebased master and remove the Rubocop related commits

Onumis:
Thanks, will release this later.

PR #281, opened 2018-10-06
Clears some rubocop offenses

harman28:
Attempts to solve #250

### DIFF
--- a/lib/rubycritic/source_control_systems/git.rb
+++ b/lib/rubycritic/source_control_systems/git.rb
@@ -21,7 +21,7 @@ module RubyCritic
       end
 
       def self.supported?
-        git('branch 2>&1') && $?.success?
+        git('branch 2>&1') && $CHILD_STATUS.success?
       end
 
       def self.to_s
@@ -37,7 +37,7 @@ module RubyCritic
       end
 
       def revision?
-        head_reference && $?.success?
+        head_reference && $CHILD_STATUS.success?
       end
 
       def head_reference
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -4,17 +4,12 @@ Metrics/AbcSize:
     - 'lib/rubycritic/analysers/smells/flay.rb'
     - 'lib/rubycritic/cli/options.rb'
 
-# Offense count: 6
+# Offense count: 2
 # Configuration parameters: CountComments.
 Metrics/MethodLength:
   Exclude:
-    - 'lib/rubycritic/analysers/helpers/ast_node.rb'
     - 'lib/rubycritic/analysers/smells/flay.rb'
-    - 'lib/rubycritic/command_factory.rb'
     - 'lib/rubycritic/core/analysed_module.rb'
-    - 'lib/rubycritic/reporter.rb'
-    - 'lib/rubycritic/revision_comparator.rb'
-    - 'lib/rubycritic/platforms/base.rb'
 
 # Offense count: 4
 Naming/AccessorMethodName:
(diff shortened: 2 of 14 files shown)


lib/rubycritic/generators/html/base.rb
```diff
@@ -45,7 +45,7 @@ def root_directory
           @root_directory ||= Pathname.new(Config.root)
         end
 
-        def get_binding
+        def binding
           binding
```
Onumis: Won't this is lead to an infinite loop?

lib/rubycritic/source_control_systems/git.rb
```diff
@@ -21,7 +21,7 @@ def git(arg)
       end
 
       def self.supported?
-        git('branch 2>&1') && $?.success?
+        git('branch 2>&1') && $CHILD_STATUS.success?
```
Onumis: This is breaking the tests. We must `require "English"` so that this works  (see https://github.com/rubocop-hq/rubocop/issues/1747).

You can add the require to the top level rubycritic module I think.

harman28: Fixing this.

> You can add the require to the top level rubycritic module
I was going to require only in the files that used that variable. Would requiring in the top module be better for some reason?

lib/rubycritic/generators/html/base.rb
harman28: Oops. Fixing.
I didn't perfectly understand the need for this function. Since it's private and can only be called in child classes, wouldn't it be just as effective to use `binding` directly in those classes instead of calling `get_binding`?

harman28: Leaving this in for now, I don't think I have enough of an understanding to change this.

harman28:
@Onumis Updated the branch. Split the changes into multiple commits for each cop fix, and also reverted some that I didn't completely understand.

Eg: `Style/MethodMissing` todo can be fixed with the following:
```
     def self.method_missing(method, *args, &block)
-      configuration.public_send(method, *args, &block)
+      if configuration.respond_to? method
+        configuration.public_send(method, *args, &block)
+      else
+        super
+      end
     end
```
But this causes the reek check to fail, with smell `method_missing manually dispatches method call`. Ideally when overriding method_missing, I believe the practice is to have a short list of methods for which we'd like special behaviour. Current code seems to want to act on all missing methods indiscriminately though.

lib/rubycritic/source_control_systems/git.rb
Onumis: Just because it might be used elsewhere. But I guess if this is the only file that uses `$CHILD_STATUS` then we can leave the require here.

lib/rubycritic/generators/html/base.rb
Onumis: Well, I'd say that we really don't need that method. I'm also not sure.
Give it a try - remove the method and see if the HTML reports are still properly generated.

lib/rubycritic/source_control_systems/git.rb
Onumis: I mean: if `$CHILD_STATUS` is only used on `RubyCritic::SourceControlSystem` then it can be imported required only in this module.

Onumis:
@harman28 as for the `method_missing` issue:
- https://github.com/troessner/reek/blob/master/docs/Manual-Dispatch.md
- https://github.com/rubocop-hq/ruby-style-guide#no-method-missing

I think we should first understand why we need a `method_missing` here.
All the config vars are defined with `attr_accessor` so there are method to read / write each one of them.

Probably we should just refactor how that configuration class works, but that should be a PR on it's own :sweat_smile:

PR #285, opened 2018-10-31
Remove some cops form .rubocop_todo.yml

AllanSiqueira:
- Removed some cops on `.rubocop_todo.yml`

Don't see the method `get_binding` being used, so I removed it, if it being used somewhere else I think we can use another approach.

If it needs some revision, I will be happy to help :)

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -1,20 +1,13 @@
 # Offense count: 1
 Metrics/AbcSize:
   Exclude:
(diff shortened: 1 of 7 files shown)


AllanSiqueira:
I'll handle that, tks : )

AllanSiqueira:
@Onumis CI problem solved :+1:

Onumis:
Thanks!

Issue #202, opened 2016-12-11
Allow using multiple formatters


thedrow:
I want to use the JSON report to generate a separate PR status for RubyCritic and I want to generate the HTML as well.
Right now if I specify `--format` twice only the last one is being picked up.

MarcGrimme:
+1 on this

JuanVqz:
It seems like this was already resolved, can we close it?

PR #288, opened 2019-02-13
Add support for multiple report formats in one run

katafrakt:
When you pass `--format` option to RubyCritic, it should be possible to use that option more than once and have reports in different formats created without having to run whole analysis twice.

My use case is to run in in CI, then extract some data from JSON report and send it to Slack, but also publish the HTML report, so people can examine it in a friendly way.

I'm not sure if tests I wrote are enough, but it was all that came to mind.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,12 @@ spec/reports
 test/tmp
 test/version_tmp
 test/samples/report.json
+test/samples/lint.txt
+test/samples/overview.html   
+test/samples/code_index.html
+test/samples/smells_index.html
+test/samples/test
+test/samples/assets
 tmp
 .idea/
 .ruby-gemset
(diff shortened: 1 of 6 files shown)


Onumis:
Mentioning #202

MarcGrimme:
Great work!! :100: 
Looks like we did nearly the same work at the same time. Funny!

MarcGrimme:
Small nit. I'd add the following to the ``.gitignore``
```
test/samples/lint.txt
test/samples/overview.html   
test/samples/code_index.html
test/samples/smells_index.html
test/samples/test
test/samples/assets
```

Onumis:
I agree. @katafrakt Can you please add those lines to gitignore, and rebase with master? Builds should be working now.

Onumis:
Wonderful!

PR #293, opened 2019-03-01
New feature to support formatters.

MarcGrimme:
This feature enables rubycritic to use external reporters chained together. A little bit similar to how simplecov does it.
One use case is a badge to be created from it like to be seen here (look on the badges):
[ruby-critic-small-badge](https://marcgrimme.github.io/rubycritic-small-badge/tmp/rubycritic/overview.html) see [rubycritic-small-badge](https://github.com/MarcGrimme/rubycritic-small-badge/)
Other use cases could be an Markdown Report but not necessarily as part of the RubyCritic gem.
Let me know what you think.

PR #294, opened 2019-03-04
Docs update

Onumis:
Follow up from #293 

@MarcGrimme made a few adjustments, what do you think?


MarcGrimme:
Makes much more sense. Thanks for taking the time.

PR #299, opened 2019-03-14
Fix generating the default html report, fixes #297

### LINKED ISSUE #297: File "tmp/rubycritic/overview.html" not being generated on v4
rubycritic v4.0.0 does not seem to generate the main overview.html report file.
ruby 2.5.3

lightalloy:
The html report was not generated in the new version (#297), cause the parsed options turned `formats` to be an empty array, so the default formats value `[:html]` was not assigned.
- fixed setting default value when `formats` are empty
- added spec for this case

### DIFF
--- a/lib/rubycritic/configuration.rb
+++ b/lib/rubycritic/configuration.rb
@@ -26,7 +26,8 @@ module RubyCritic
     end
 
     def setup_formats(options)
-      self.formats = options[:formats] || [:html]
+      formats = options[:formats].to_a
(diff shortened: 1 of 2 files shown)


Onumis:
Great! Think this is related to #297 which I totally forgot about since I reported that!

lightalloy:
@Onumis yep, I referenced this issue in the description ) Thanks for merging!

Onumis:
Thank you for pushing the fix! It's live now.

PR #305, opened 2019-03-25
Fixing a css typo

ut:
I've fixed a small typo at application.css, which will break SCSS conversion. This fix relates to the current release of Rubycritic 4.0.2


Onumis:
Thanks!

PR #306, opened 2019-04-04
Config options getting overwritten by nil argv options

nathanbwright:
Options that are defined in the config file were getting overwritten by everything defined by the `argv_options` hash because of the way the hashes are getting merged here: https://github.com/whitesmith/rubycritic/blob/master/lib/rubycritic/cli/options.rb#L23

This commit deletes all argv options that are nil which allows any config file options to be applied as expected.

Another option would be to supply a block to the hash merge referenced in the above link. I initially went this route but then realized that you can't simply check for truthy-ness because the argv options could be false...

### DIFF
--- a/lib/rubycritic/cli/options.rb
+++ b/lib/rubycritic/cli/options.rb
@@ -24,9 +24,9 @@ module RubyCritic
         file_hash = file_options.to_h
         argv_hash = argv_options.to_h
 
-        file_hash.merge(argv_hash) { |_, file_option, argv_option|
+        file_hash.merge(argv_hash) do |_, file_option, argv_option|
           argv_option.nil? ? file_option : argv_option
-        }
+        end
       end
     end
   end
--- a/lib/rubycritic/cli/options.rb
+++ b/lib/rubycritic/cli/options.rb
@@ -19,6 +19,7 @@ module RubyCritic
         self
       end
 
+      # :reek:NilCheck
       def to_h
         file_hash = file_options.to_h
         argv_hash = argv_options.to_h
--- a/lib/rubycritic/cli/options.rb
+++ b/lib/rubycritic/cli/options.rb
@@ -20,7 +20,12 @@ module RubyCritic
       end
 
       def to_h
-        file_options.to_h.merge(argv_options.to_h)
+        file_hash = file_options.to_h
+        argv_hash = argv_options.to_h
+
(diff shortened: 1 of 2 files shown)


nathanbwright:
Sorry, I must not have explained myself very well. Options (such as `paths`) that are defined in in the config file are being overwritten by everything and anything that is defined in the `argv_options` even if it is the option has a nil value, because of the way merging hashes works (without a block).

I've added some explanatory variables.

```
[22, 31] in /Users/home/.rvm/gems/ruby-2.5.3@gemset/gems/rubycritic-4.0.1/lib/rubycritic/cli/options.rb
   22:       def to_h
   23:         file_options_hash = file_options.to_h
   24:         argv_options_hash = argv_options.to_h
   25:         merged_options = file_options_hash.merge(argv_options_hash)
   26:         byebug
=> 27:         merged_options
   28:       end
   29:     end
   30:   end
   31: end
(byebug) file_options_hash
{
# ...
 :paths=>["app/", "lib"],
# ...
}

(byebug) argv_options_hash
{
# ...
 :paths=>nil,
# ...
}

(byebug) merged_options
{
# ...
 :paths=>nil,
# ...
}
```

Do you see what I'm saying?

Onumis:
Ah yes. Very well then, your solution should do the trick. Just make sure travis is ok with it as well and I'll merge this in! Thanks.

nathanbwright:
It looks like `reek` isn't happy because of the explicit nil check. We can't use a normal truthy-ness check in this case because `false` is a valid option for the `argv_options`. 

Should I just add a comment to suppress the `reek` warning? Or would you like a different solution?

Onumis:
I'd this can be set to be ignored by reek.

nathanbwright:
I'm thinking it makes more sense to put the hash merging logic into the main options class instead of "magically" removing keys from the argv options class that had nil values... What do you think @Onumis?

lib/rubycritic/cli/options.rb
```diff
@@ -19,8 +19,14 @@ def parse
         self
       end
 
+      # :reek:NilCheck
```
Onumis: Does this disable the NilCheck for the method, to the end of file, or "from here on" ?

nathanbwright: https://github.com/troessner/reek/blob/master/docs/Smell-Suppression.md

Onumis:
Thanks! I'll try to make a release asap.

PR #309, opened 2019-06-07
Update Reek dependency

leksster:


Onumis:
I was thinking that maybe this dependency requirement is too much constrained. 
Since reek follow semantic versioning, wdyt of changing to `'reek', '~> 5.0', '< 6.0'`?

leksster:
Completely agree. Thanks.

PR #315, opened 2019-08-02
fix typo in example `.rubycritic.yml`

Adre:
[ci skip]

Using this example as a starting point for config results in `threshold_score` being 0 as the key is incorrect.


Onumis:
Nice!

Issue #245, opened 2017-07-26
Feature Request: Consume SimpleCov and Use Results to Weigh Metrics


TreyE:
It would be great if I could somehow also import simplecov results into rubycritic - and then as a result use those coverage values to help 'weigh' my metrics.  Basically I want lack of test coverage to act as something that magnifies how 'bad' I consider the code to be.

A simple case would be:

1. Prior to the application of the coverage metric, both classes A and B receive a the same rubycritic score
2. Class A has full test coverage
3. Class B has no test coverage
4. Class A should either receive an improved rating or keep it's current rating.
5. Class B should receive a far worse rating that it would have prior to the code coverage metric being factored.

Onumis:
@TreyE alright so the flow would be:

- run your test suite(s) with codecov
- run rubycritic passing in the codecov `.resultset.json`
    - for each examined file, look up the code coverage results for that file
    - assign a score based on it, which will affect the final rating

Although I don't think we should mix code quality with test coverage, perhaps this is useful for somebody.
Want to get started with a PR? :D

gonzaloaune:
👍

joshsaintjacque:
@TreyE Is this something you're working on? If not, I'd love to make a PR.

TreyE:
Nope, nothing I've been working on - just something I hope somebody picks up.

etagwerker:
@Onumis @TreyE @joshsaintjacque I really like this idea. I like it so much that I submitted a PR for it: #319 

I'd really like to know your feedback about those changes. 

Coverage and quality data are not mixed nor combined. Coverage data now shows up in a new section. 

There is no need to pass in the coverage folder or `.resultset.json`, the changes I propose use SimpleCov's configuration to find that file. 😸

etagwerker:
In case you are interested in this feature, I recently release [`skunk`](https://github.com/fastruby/skunk) which uses `rubycritic` and `simplecov` to calculate **one metric** called the "StinkScore" 

I wrote a blog post about this: https://www.fastruby.io/blog/code-quality/intruducing-skunk-stink-score-calculator.html 😃

PR #319, opened 2019-10-10
SimpleCov Support

etagwerker:
Hey @Onumis @TreyE and friends! 

This PR is an attempt to solve a part of https://github.com/whitesmith/rubycritic/issues/245

For now it just lists files and its coverage in a new section. It also addresses an item in the [ROADMAP.md](ROADMAP.md): 

>  Integrate other gems, like:
> 
>  Simplecov to provide code coverage

Before I work more on this, I figured I could send the PR and get your feedback. A couple of questions that I had: 

1. Do you think it is a good idea to add it to RubyCritic's core? 
2. Or, do you think it should be a separate gem? (e.g. rubycritic-simplecov) 

A lot of the code that is in `lib/rubycritic/analysers/coverage.rb` comes from SimpleCov's merger: https://github.com/colszowka/simplecov/blob/master/lib/simplecov/result_merger.rb#L14-L115 (sorry I didn't find a classier way) 

Here is what it looks like: 

ruby-critic-coverage

Compared to SimpleCov's report: 

simplecov-coverage-gggiiifff-com

This is the sample application that I used: https://github.com/fastruby/gggiiifff.com

Please check it out and let me know if I am on the right track. 😄 

Thanks!


etagwerker:
@Onumis @TreyE just fixed the failing specs. Let me know what you think.

lib/rubycritic/analysers/coverage.rb
```diff
@@ -0,0 +1,117 @@
...
+      end
+
+      # Loads the cached resultset from JSON and returns it as a Hash,
+      # caching it for subsequent accesses.
+      def resultset
+        @resultset ||= begin
+          if (data = stored_data)
+            begin
+              JSON.parse(data) || {}
+            rescue
```
Onumis: This should rescue from `JSON::ParserError` or at least `StandardError` and probably print to STDOUT.
It should not happen, but if it does the user would like to know.

etagwerker:
@Onumis Just addressed the issues you mentioned. Please let me know if it is good now.

etagwerker:
@Onumis Did you see anything else that needs to be improved? Just curious... 😸

Onumis:
LGTM! 🎉 
Will merge it now and prepare a release later today. 
Thanks! 🙏

silvermind:
Hi @etagwerker nice PR 👍 
As we're using special coverage dir in the `rails_helper.rb` for the Specs (coverage_dir ENV.fetch('SIMPLE_COV_OUTPUT_DIR')), we don't get any results in rubycritc context.
As this config is only loaded in rspec context, moving the same to the `development.rb` or `test.rb` environments didn't help too, as rubycritic does not load them at all.. any ideas how to solve this in the rubycritic context?

etagwerker:
@silvermind I'm not sure what could be wrong there. I wrote the PR so that it uses `SimpleCov.coverage_dir` when trying to find the `.resultset.json` file. Could you open an issue here? https://github.com/whitesmith/rubycritic/issues/new

That way we can continue the conversation over there.

PR #328, opened 2019-10-30
Update details in ROADMAP.md 

etagwerker:
Hey @Onumis, 

I think it might be useful to update the `ROADMAP.md` based on things that are already present in v4.2.0

What do you prefer? Marking items as done or removing them from ROADMAP.md?

I decided to mark these items as done because: 

## Project Score 

It is already present in the console: 

Screen Shot 2019-10-30 at 3 25 00 PM

Screen Shot 2019-10-30 at 3 25 10 PM

## SimpleCov

Added in #319 

Let me know.

Thanks,
Ernesto


Onumis:
Yep, checking them out is fine. Thanks!

Issue #277, opened 2018-09-12
Exit code 1 due to uncommitted changes


rlgreen91:
I'm running Rubycritic as part of a series of rake tasks within CodeClimate.  Ironically enough, I have a PR upgrading Rubycritic to the latest version.  However, CI keeps failing with the following error:

```
bundle exec rubycritic --minimum-score 85 --mode-ci --no-browser --path /tmp/artifacts/rubycritic app config lib
Uncommitted changes are present.
Exited with code 1
```

Has anyone seen this error before?  I'm trying to figure out where this is coming from, since all tracked files in the branch have been committed and running rubycritic locally completes successfully.  My Ruby version is 2.5

Onumis:
Hmm, must have been introduced here: https://github.com/whitesmith/rubycritic/pull/271
:thinking:

Onumis:
Reviewing the method `uncommitted_changes?`:

```ruby
      def self.uncommitted_changes?
        !`git ls-files --other --exclude-standard --directory`.empty? || !`git status --porcelain`.empty?
      end
```

I think we need to drop the `git ls-files` part, since it seems it will output folders like `tmp`, which are not tracked but might exist after running rubycritic, or simply loading a rails app.

`git status --porcelain` might be enough to tell if there are uncommitted changes, or even a `git diff-index HEAD --`

Will read on this and see what is best: https://stackoverflow.com/questions/3878624/how-do-i-programmatically-determine-if-there-are-uncommitted-changes

rlgreen91:
I wanted to come back around and explain what exactly the issue was for us.  So, we encountered this issue running Rubycritic on CircleCI.  The thing about CircleCI and other CI services is that we don't have any of the gems installed - it has to do that.  As part of that build script, we would create a `/vendor` directory that CircleCI would install the gems in.  Rubycritic would then complain about uncommitted changes, which occurred because that directory literally didn't exist until CircleCI started running.  So that was our issue.  Adding `/vendor` to our .gitignore solved it for now.

Onumis:
Thanks for the info! Anyway, the fix I proposed is still a better way of detecting such things.

PR #331, opened 2019-12-05
Show what's uncommitted

GeoffTidey:
Thought I'd make this into a PR as it helped me discover what files are actually uncommitted on our CI server, and it may help someone else from wasting a few hours...

(related to https://github.com/whitesmith/rubycritic/issues/277)

### DIFF
--- a/lib/rubycritic/source_control_systems/git.rb
+++ b/lib/rubycritic/source_control_systems/git.rb
@@ -56,7 +56,7 @@ module RubyCritic
       end
 
       def self.uncommitted_changes
-        git('diff-index HEAD --').chomp!
+        git('diff-index HEAD --').chomp! || ''
       end
 
       def self.uncommitted_changes?
--- a/lib/rubycritic/source_control_systems/git.rb
+++ b/lib/rubycritic/source_control_systems/git.rb
@@ -60,7 +60,7 @@ module RubyCritic
       end
 
       def self.uncommitted_changes?
-        uncommitted_changes.present?
+        !uncommitted_changes.empty?
       end
 
       def self.modified_files
--- a/lib/rubycritic/source_control_systems/git.rb
+++ b/lib/rubycritic/source_control_systems/git.rb
@@ -60,7 +60,7 @@ module RubyCritic
       end
 
       def self.uncommitted_changes?
-        !uncommitted_changes.empty?
+        uncommitted_changes.present?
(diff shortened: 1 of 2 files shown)


lib/rubycritic/source_control_systems/git.rb
```diff
@@ -52,11 +52,15 @@ def travel_to_head
       end
 
       def self.switch_branch(branch)
-        uncommitted_changes? ? abort('Uncommitted changes are present.') : `git checkout #{branch}`
+        uncommitted_changes? ? abort("Uncommitted changes are present: #{uncommitted_changes}") : `git checkout #{branch}`
```
Onumis: This can be changed into the following to avoid big lines:

```ruby
dirty = !uncommitted_changes.blank?
abort("Uncommitted changes are present: #{uncommitted_changes}") if dirty? 

`git checkout #{branch}`
```

Onumis: `self.uncommitted_changes?` can then be deleted, or change `self.uncommitted_changes` so that it memoizes the result (because calling git via an OS call is expensive): 

```ruby
def self.uncommitted_changes
    return @uncommitted_changes if defined? @uncommitted_changes
    @uncommitted_changes = git('diff-index HEAD --').chomp! || ''
 end
```

(from the top of my head I don't think it will be a problem to have this value cached 🤔 )

lib/rubycritic/source_control_systems/git.rb
```diff
@@ -52,11 +52,16 @@ def travel_to_head
       end
 
       def self.switch_branch(branch)
-        uncommitted_changes? ? abort('Uncommitted changes are present.') : `git checkout #{branch}`
+        dirty = !uncommitted_changes.blank?
+        abort("Uncommitted changes are present: #{uncommitted_changes}") if dirty
+
+        `git checkout #{branch}`
```
Onumis: While we're at it, can this be changed so that it also uses the `self.git` method? `git "checkout #{branch}"` 🙏

Onumis:
```suggestion
        git "checkout #{branch}"
```

GeoffTidey:
Cheers @Onumis, changes made

GeoffTidey:
Hey @Onumis, done 👍

PR #337, opened 2019-12-27
Fix `switch_branch': undefined method `blank?' for "":String (NoMethodError)

eitoball:
This PR should fix an error like following when running version 4.3.0 of rubycritic.

```
../rubycritic-4.3.0/lib/rubycritic/source_control_systems/git.rb:55:in `switch_branch': undefined method `blank?' for "":String (NoMethodError)
```

`String#blank?` is defined by [ActiveSupport](https://api.rubyonrails.org/classes/String.html#method-i-blank-3F), not in ruby standard `String` class.

Check list:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # master [(unreleased)](https://github.com/whitesmith/rubycritic/compare/v4.3.0...master)
 
+* [BUGFIX] Fixes NoMethodError on RubyCritic::SourceControlSystem::Git.switch_branch (by [@eitoball][])
+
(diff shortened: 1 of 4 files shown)


nunosilva800:
#338

nunosilva800:
Thanks for the quick fix! 
I am not sure why the tests are failing on some ruby versions and other not, but it seems that the new mock introduced is "leaking" into other tests... Any idea on why that happens?

eitoball:
@nunosilva800 

I think that I found the cause. `RubyCritic::SourceControlSystem::Git.switch_branch` was `class_eval`-ed in `test/lib/rubycritic/commands/compare_test.rb`. Depending on execution order of tests, it looked like leaking.

Issue #321, opened 2019-10-16
Flay | How to disable it or exclude some files from its analysis?


lflucasferreira:
Hi,

The Flay is pointing a DuplicateCode Smell in two files. I would like to ignore these files and increase the code score, but there is no documentation about how to disable/exclude this smell.

Could you please help me? I'm lost.

Yinfei:
Hello 👋 

I tried introducing a `.flayignore` file but it seems that it's ignore by rubycritic itself. 🤔 

Is there any way that rubycritic could take this into account?

mvz:
@Yinfei sorry, I'm not very familiar with how rubycritic works. I only work on the Reek project.

lflucasferreira:
> Hello 👋
> 
> I tried introducing a `.flayignore` file but it seems that it's ignore by rubycritic itself. 🤔
> 
> @mvz Is there any way that rubycritic could take this into account?

Could you please tell me how to configure the `.flyignore?` file? TYIA.

mvz:
I also don't know anything about Flay, which is a separate project.

PR #348, opened 2020-02-12
Take .flayignore file into account

Flink:
Flay accepts a `.flayignore` file containing paths allowing to skip its
analysis for the provided paths. The current version of Rubycritic
ignores this file and Flay will always apply to all the paths provided
to Rubycritic.

This patch addresses this issue by taking a small snippet of code from
the Flay source itself and apply it to the Flay helper class. Now the
`.flayignore` file is honored as expected.


Flink:
This should resolve https://github.com/whitesmith/rubycritic/issues/321 :)

nunosilva800:
Very nice, thanks a lot!

Issue #37, opened 2015-03-16
Allow configuring date range of Churn calculation


natesholland:
I'm running rubycritic on a large project that's been running for 10 years so our git log is very long. I'm thinking we should limit it by commits within the last year in order to speed things up. My calculations showed that this would speed up the churn section by 500% on older projects.

bglusman:
Relatedly @guilhermesimoes, I was just trying to upgrade debt_ceiling to the new 1.4 version so I can avoid using the rubycritic internals the way you showed me a few months back, but it's WAY slower now, even in :ci mode, which I beleive skips the churn analysis, correct?  I'd like to use your latest, but I may have to roll back to old version unless there's some undocumented way I'm missing to get the speed I was getting in [this code](https://github.com/bglusman/debt_ceiling/blob/v0.1.1/lib/debt_ceiling/accounting.rb#L39)

guilhermesimoes:
@bglusman `--mode-ci` does not skip churn calculation, since we need it to generate that "Complexity vs Churn" scatter plot. That mode simply makes sure that RubyCritic runs once instead of twice.

The things is, using the new API, you're running [all these analysers](https://github.com/whitesmith/rubycritic/blob/4219403a524db249ebd18d00ba8e836e9f486b5f/lib/rubycritic/analysers_runner.rb#L14-L19). So obviously it will be slower than just the two you were using previously. I'll have to work on something to let you pick which analysers to run...

bglusman:
@guilhermesimoes hmm, OK, that makes sense...  after some changes I'm actually using most of all of these analyzers now, EXCEPT for churn, so being able to turn that off (which I thought we discussed specifically for ci mode in November) would be great, but yeah, making it a more general feature to specify which makes sense.  I suppose since I'm using most of them now I could leave it as is, or go back to semi-manual/hacky user of your internal API until that's done.  I could probably do a PR to make that change myself this week if you'd like, or let me know if you prefer to do yourself?

PR #350, opened 2020-02-19
Make the revision count faster

tomchipchase:
Changed the revision count from running git log for every file to run
git log once for the project and process the results.

This solves a problem for us trying to add rubycritic to our CI pipeline where it would take 40 minutes to run. After this change it takes 5 minutes, and the churn stage itself is done in seconds.

This indirectly fixes https://github.com/whitesmith/rubycritic/issues/37 by making the churn fast enough that it would no longer be necessary to filter commits for speed.

Check list:

### DIFF
--- a/lib/rubycritic/source_control_systems/git/churn.rb
+++ b/lib/rubycritic/source_control_systems/git/churn.rb
@@ -1,0 +1,87 @@
+# frozen_string_literal: true
+
+module RubyCritic
+  module SourceControlSystem
+    class Git < Base
+      Stats = Struct.new(:count, :date)
+
+      class Renames
+        def initialize
+          @data = {}
+        end
+
+        def renamed(from, to)
+          current = current(to)
+          @data[from] = current
+        end
+
+        def current(name)
+          @data.fetch(name, name)
+        end
+      end
+
+      class Churn
+        def initialize
+          @renames = Renames.new
+          @date = nil
+          @stats = {}
+
+          call
+        end
+
+        def revisions_count(path)
+          stats(path).count
+        end
+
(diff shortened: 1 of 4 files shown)


lib/rubycritic/source_control_systems/git/churn.rb
```diff
@@ -0,0 +1,89 @@
...
+          stats(path).date
+        end
+
+        private
+
+        def call
+          Git
+            .git("log --all --follow --format='format:date:\t%ad' --name-status .")
+            .split("\n")
+            .reject { |line| line =~ /^$/ }
```
nunosilva800: So this removes empty lines?  Might be more readable to use `.reject(&:empty?)`

lib/rubycritic/source_control_systems/git/churn.rb
```diff
@@ -0,0 +1,89 @@
...
+            process_date(rest)
+          when /^R/
+            process_rename(rest)
+          else
+            process_file(rest)
+          end
+        end
+
+        def process_date(date)
+          @date = DateTime.parse(date)
```
nunosilva800: Since the goal is to optimise the speed, maybe the date can be stored as a string, and only parsed+memoized when `date_of_last_commit` is called? Or as a last step of `call`.

The methods that rely on `Stats` might need to be updated as well. I'll leave it to your to see if it's worth it or not :)

tomchipchase: After reviewing the original code, it looks like it doesn't parse the date at all, and leaves it as a string, so that would make it match the original. I'll remove the parsing.

Issue #166, opened 2016-10-07
Reek configuration not being picked up


jgrau:
With a fresh `gem install` of `rubycritic` and a run with `rubycritic --no-browser --minimum-score 85 --path critique app lib config script test` in my rails project I get a score of 89.72. 

Running `reek --todo` which correctly generates a `.todo.reek` file with a lot of ignores, I would expect the subsequent run of `rubycritic` would give me a higher score. However, I get the exact same score. Checking the report I see a lot of smells which should have been ignored.

```
|            | Version |
|------------|---------|
| Ruby       | 2.2.3   |
| RubyCritic | 2.9.4   |
| Reek       | 4.4.1   |
```

mereghost:
It seems that the instantiate a `Reek::Examiner` without any options. That loads a default configuration using `Reek::Configuration::AppConfiguration.default` but that does not trigger the config file loading.

Reek's CLI passes to the examiner the result of `Reek::Configuration::AppConfiguration.from_path(config_file)` for the config file handling.

troessner:
Yeah, that's most certainly a bug. Pull requests welcome!

mereghost:
Any contribution guidelines that I must be aware of before sending you folks a PR?

troessner:
Check out [Contributing](https://github.com/whitesmith/rubycritic/blob/master/CONTRIBUTING.md)

joneshf:
I see there was a PR for this bug (https://github.com/whitesmith/rubycritic/pull/175), but it seems to still be a problem:

```shell
$ cat .reek
---
exclude_paths:
  - lib/templates
$ reek lib/templates      
Inspecting 0 file(s):


$ rubycritic lib/templates
  lib/templates/active_record/model/model.rb:1 :: parse error on value "<" (tLT)
  skipping lib/templates/active_record/model/model.rb
running flay smells

running flog smells
[...]

PR #352, opened 2020-03-02
Reek excluded_path configuration not being picked up

AllanSiqueira:
## Motivation

When using the key `excluded_files` in `.reek.yml`, the parameter have been ignored in the rubycritic

## Proposal

Looking in the Reek implementation, this filter is applied before he comes in the `Reek::Examiner`, so, a workaround for this issue is to filter the files before sending to Reek::Examiner.

I can add more test scenarios for corner cases if you can give me insights about this flow.

## Check list:
- [X] Add an entry to the [changelog](/CHANGELOG.md)
- [X] [Squash all commits into a single one](/CONTRIBUTING.md)

Issue: #166

### DIFF
--- a/lib/rubycritic/analysers/smells/reek.rb
+++ b/lib/rubycritic/analysers/smells/reek.rb
@@ -13,7 +13,7 @@ module RubyCritic
       end
 
       def run
-        @analysed_modules.each do |analysed_module|
+        relevant_paths.each do |analysed_module|
           add_smells_to(analysed_module)
           print green '.'
         end
@@ -25,6 +25,8 @@ module RubyCritic
       end
 
       private
+
+      attr_reader :analysed_modules
 
       def add_smells_to(analysed_module)
         Reek.new(analysed_module.pathname).smells.each do |smell|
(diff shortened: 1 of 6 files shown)


lib/rubycritic/analysers/smells/reek.rb
```diff
@@ -26,6 +26,8 @@ def to_s
 
       private
 
+      attr_reader :analysed_modules
+
       def add_smells_to(analysed_module)
         Reek.new(analysed_module.pathname).smells.each do |smell|
```
nunosilva800: Maybe by using `Reek::Examiner` instead of the top level module will work?

`examiner = Reek::Examiner.new(source)`

Also, this `examiner` can be memoized.

AllanSiqueira: This is not using the top level module, he is using the `lib/rubycritic/analysers/helpers/reek.rb` module, I tried use the examiner this way, passing the configuration and still dont work

Doing some research on reek's source code I see the `excluded_paths` param is filtered before he reach the `Reek::Examiner`, in [SourceLocator](https://github.com/troessner/reek/blob/master/lib/reek/source/source_locator.rb).

PR #353, opened 2020-03-15
Update rubocop to the latest version

MiralDesai:
Just a small PR to update rubocop to the current latest version. Was surprised to be a very old version being used here so decided to submit this.

Check list:
- [✅] Add an entry to the [changelog](/CHANGELOG.md)
- [✅] [Squash all commits into a single one](/CONTRIBUTING.md)
- [✅] Describe your PR, link issues, etc.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -30,4 +30,7 @@ Style/RedundantFreeze:
     - 'lib/rubycritic/source_locator.rb'
     - 'lib/rubycritic/version.rb'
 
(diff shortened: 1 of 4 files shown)


nunosilva800:
Thanks for the PR! In fact rubocop is only a development dependency for rubycritic - it is not used to generate reports.
Nevertheless its nice to keep dependencies up to date. This change though, conflicts with Reek, as it does not want `UncommunicativeVariableName`.

Can you update this please?

Issue #287, opened 2019-01-14
Report directory structure when comparing branches seems off


dsager:
When running rubycritic in CI mode, the report-folder seems to be weirdly nested.

After executing the following command (while on a branch called `test-123`):
```
$ rubycritic --mode-ci=develop --format=json app
```
I get the following directory structure:
```
tmp/rubycritic/
├── build_number.txt
└── compare
    ├── build_details.txt
    └── develop
        ├── compare
        │   └── test-123
        │       ├── compare
        │       │   └── report.json
        │       └── report.json
        └── report.json
```
I would expect the structure to be this:
```
tmp/rubycritic/
├── build_number.txt
└── compare
    ├── build_details.txt
    ├── develop
    │   └── report.json
    ├── test-123
    │   └── report.json
    └── report.json
```

Actually, when generating a HTML report, the links in the top right corner also seem to expect this structure.

|            | Version |
|------------|---------|
| Ruby       | 2.3.7   |
| RubyCritic | 3.5.2   |

I am happy to create a PR for this, jut wanted to confirm first if this is an actual issue. I think the problem is that `Config.root` is overwritten in two places during the branch comparison and not reset to its initial value: [here](https://github.com/whitesmith/rubycritic/blob/v3.5.2/lib/rubycritic/commands/compare.rb#L52) and [here](https://github.com/whitesmith/rubycritic/blob/v3.5.2/lib/rubycritic/commands/compare.rb#L60).

While at it, I have two additional questions regarding the CI mode (happy to create separate issues if preferred):
- The help (and also the readme) states that the CI mode will be faster. This currently seems to not be the case, as it runs a full analysis of both branches, and then looks at the diff. So it's about twice as slow as the normal run. Maybe I am doing something wrong (see command above)?
- Wouldn't it make more sense to include the information stored in `build_details.txt` in the `json` output, so that it can be easily parsed?

GeoffTidey:
I can confirm (one year on) that I'm seeing the same issue as @dsager

We're using ruby 2.7 & rubycritic 4.3.1 &

`rubycritic app -p $CIRCLE_ARTIFACTS/rubycritic --mode-ci -f html --no-browser `

An example of an actual file on disk is:
`/tmp/circleci-artifacts/rubycritic/compare/master/compare/staging/overview.html`

whilst the link in the top right corner is pointing at:

`/tmp/circleci-artifacts/rubycritic/compare/staging/overview.html`

denny:
[...]

PR #378, opened 2021-01-21
Fix compare file structure

denny:
This PR fixes issue #287; the reports for both branches and the diff report all end up where the top-of-page links expect them to be (which is also where humans expect them to be, going by the comments on that issue).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * [CHANGE] Fix rubocop offenses (by [@sl4vr][])
 * [CHANGE] Make Github Linguist ignore vendored files (by [@sl4vr][])
+* [BUGFIX] Fix directory structure of reports when comparing branches (by [@denny][])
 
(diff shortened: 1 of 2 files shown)


denny:
@etagwerker, as you commented on the issue, perhaps you could review this when you get a chance?

I went for a minimal intervention approach here - just stashing a copy of the root path before it starts being changed, and using that copy as the 'source of truth' in the method that was concatenating successive changes before.

I think a more correct solution would be not to change Config.root_path at all - it doesn't sound like something that should be mutable - but fixing that looked like it would involve changing a lot more code.

etagwerker (commented):
@denny Looks good 👍 

Could you please update the CHANGELOG as well?

denny:
Oh yeah, sorry, I meant to ask about that; what do I do about version numbers for the changelog, am I meant to bump the patch level in the PR and use the new number?

etagwerker:
@denny if you can add a note that's fine. I'll bump the version right before I cut a new release.

For reference you can check some of the latest PRs we have merged. 

Thank you! ❤️

etagwerker (approved):
@denny Great, thanks! 💯

PR #380, opened 2021-01-26
Prepare v4.5.3

nunosilva800:


etagwerker (approved):
@nunosilva800 Looks good to me 👍 

The only question I have is: Do you think it is a good idea to release 4.5.3 instead of 4.6.0? I'm asking because https://github.com/whitesmith/rubycritic/pull/378 might break the "public interface" for this gem. I'm not sure how many people have code out there that might rely on the previous structure for directories...

nunosilva800:
Hmm, good point. That probably warrants a v4.6.0. Thanks.

PR #372, opened 2020-10-03
Fix issue 307 comparing files on two branches

lauratpa:
fixes #307 
This bug occurs when a branch that is being compared contains a file that is not present on the other branch.
In this template the files are being compared to one another, but not all files have to have corresponding file on the second branch.

I would have wanted to add a test for this bug, but there seem to be no existing
tests for the template files.

PR #382, opened 2021-01-26
Add basic test for HTML report generation

denny:
This is a naive copy-and-adapt of the existing `json_report_test.rb`; I created it to see if I could work up a test for #372, but I think I'm going to have to concede defeat on that - there's too much mocking going on that I don't understand properly - but I thought this file might at least be a starting point for somebody else to use as a base for adding that test in the future?

Let me know if this looks useful, and if so I'll squash it and do the CHANGELOG thing - probably after you do your imminent release.

### DIFF
--- a/test/lib/rubycritic/generators/html_report_test.rb
+++ b/test/lib/rubycritic/generators/html_report_test.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+require 'rubycritic/analysers_runner'
+require 'rubycritic/generators/html_report'
+require 'fakefs_helper'
+
+describe RubyCritic::Generator::HtmlReport do
+  describe '#generate_report' do
+    around do |example|
+      capture_output_streams do
+        with_cloned_fs(&example)
+      end
+    end
+
+    it 'creates the HTML report files' do
+      sample_files = Dir['test/samples/**/*.rb']
+      create_analysed_modules_collection
+      generate_report
+      assert File.file? 'test/samples/overview.html'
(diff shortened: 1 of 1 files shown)


denny:
Seems to be failing on Ruby 2.5 and 2.6 (but not 2.4, wut??) on Travis CI - again, at first glance I don't think I know what's going on there, sorry.

Least useful PR ever 😄

denny:
Oh, except... I just ran the test suite under 2.7.2 on my machine and it failed with the same error the other two have on Travis. Then I blew away all the things that are ignored in the .gitignore - tmp, the output in test/samples, etc etc - and the tests ran successfully.

Then I switched to 2.6.6 and the tests ran okay without clearing temp files again (?!)

Then 2.5.8, where they had the same error, which again cleared when I cleared the temp files.

I still have no idea what's going on, but maybe somebody else can guess based on that lot.

denny:
In vaguely related news, the tests fail early for me on Ruby 3.0.0, with "File does not exist: rexml/document"

Presumably related to this in 3.0.0 release notes:
> The following default gems are now bundled gems.
> * rexml

I'm not sure what adding it to the Gemfile would do to older Rubies, so I haven't.

nunosilva800:
Thanks for this investigation @denny I'll have a better look towards the weekend.
Regarding ruby3, we should add that to the travis config, and allow failures for now. Then create issues to address those failures and prepare a release that supports ruby3. :)

etagwerker:
Continuing this work over here: https://github.com/whitesmith/rubycritic/pull/483

PR #385, opened 2021-01-28
Add Support for Ruby 3.0 

etagwerker:
Hi there, 

This is an attempt to make RubyCritic compatible with Ruby 3.0. 

It seems that we will have to change a few dependencies: 

- SimpleCov 
- Rexml (no longer comes with Ruby) 
- Minitest 

Check list:

- [x] Add an entry to the [changelog](/CHANGELOG.md)
- [x] [Squash all commits into a single one](/CONTRIBUTING.md)
- [x] Describe your PR, link issues, etc.
- [x] Get test suite to pass with Ruby 3.0 

Please check it out and let me know what you think.

Thanks!

PR #392, opened 2021-03-26
CI: Drop rbx-3 from matrix, does not install

olleolleolle:
This PR changes the CI matrix by removing a not-installable Ruby version (rbx-3).

Inspired by #385.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,9 @@ jobs:
     - rvm: ruby-head
     - rvm: jruby-9.1.17.0
       env: JRUBY_OPTS='--debug' # get more accurate coverage data
-    - rvm: rbx-3
   allow_failures:
     - rvm: ruby-head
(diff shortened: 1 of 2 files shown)


etagwerker (approved):
@olleolleolle Looks good to me. I'm curious: Why doesn't it install? Is there a known issue or a problem with this project's dependencies?

olleolleolle:
@etagwerker TravisCI and the Rubinius project both no longer put out versions that can install. 

Travis in the old days, would keep up, and make sure there were just-name-it-and-it-will-be-there versions. Later on, that fell away. 

And then later, the Rubinius project slowed down. 

Now the project no longer calls itself a Ruby but a language framework/platform.

So, I think it's fair to say that none of rubycritic's code/deps have anything to do with the "not installable".

etagwerker:
@olleolleolle thanks for sharing more info! 👍

Issue #308, opened 2019-05-03
Ignores Ruby files without a .rb extension


cvoltz:
Rubycritic won't analyze files which don't have the `.rb` suffix even if the file has a shebang line like:
```
#!/usr/bin/env ruby 
```
which indicates it it is a Ruby file. This is true even if the file is explicitly specified either as a trailing argument on the CLI:
```
rubycritic exe/my-bin
```
or by using the `paths` argument in the Rake task:
```
RubyCritic::RakeTask.new do |task|
  task.paths = FileList["exe/my-bin"]
  task.verbose = true
end
```

The problem seems to be that the code checks to see if the file has the '.rb' extension: https://github.com/whitesmith/rubycritic/blob/a6ddc92235adb2f4353dfec1286850ebbe48a800/lib/rubycritic/source_locator.rb#L42

This is unnecessarily restrictive. When creating a gem which provides a command the user will run from the CLI, the convention is to provide the Ruby script in the `exe` directory (which the `gem` will use to put the script into a `bin` directory in the user's `PATH`), add a shebang to run the Ruby interpreter, and set the executable permission on the file (i.e., `chmod +x exe/my-bin`). Rubycritic doesn't provide any means to analyze such a script.

As an ugly workaround, I have to create a symlink (with the `.rb` suffix) for each Ruby script in the `exe` directory and make sure the `.gemspec` file excludes the symlinked files.

A good solution would be to analyze all files specified explicitly (on the CLI or via the `paths` argument of the Rake task) and to look at the shebang line of the files which were globbed. Rubocop does this already so you can look at the signatures they look for in the shebang to decide if the file is a Ruby file or not.

etagwerker:
@cvoltz Thanks for reporting this! I like your solution, I think we need to improve the way we detect whether it is a file with Ruby code or not. 

In terms of implementing a patch, I would recommend we go with a solution like this: 

1. First check for the extension. If the extension is .rb then it is a Ruby file -- This already works
2. Then check for the first line in the file, if it has the ruby shebang, then it is a Ruby file. -- This needs to be implemented. 

If you want to take a stab at it, I'd be happy to review your PR. 👍

NerdyBoyCool:
@etagwerker 
If no one else is working on this yet, can I work on it?
I plan to work on this issue using rubocop's FileFinder module as a reference.

etagwerker:
@NerdyBoyCool  as far as I know nobody is actively working on this issue so I'd say go for it. 👍

[...]

PR #394, opened 2021-05-07
Include files which have a ruby shebang

stufro:
Check list:

-----
Issue #308 

Implemented by checking the first line of the file for a ruby shebang, it will only check for shebang if the file does not have a .rb file extension.

I added the new private method I created to the .reek.yml to ignore the UtilityFunction warning. I couldn't think of a way to do this that didn't seem like overkill, I'm open to ideas if there's a better way.

### DIFF
--- a/lib/rubycritic/source_locator.rb
+++ b/lib/rubycritic/source_locator.rb
@@ -7,6 +7,7 @@ module RubyCritic
   class SourceLocator
     RUBY_EXTENSION = '.rb'.freeze
     RUBY_FILES = File.join('**', "*#{RUBY_EXTENSION}")
+    RUBY_SHEBANG = '#!/usr/bin/env ruby'.freeze
 
     def initialize(paths)
       @initial_paths = Array(paths)
@@ -39,7 +40,7 @@ module RubyCritic
       path_list = @initial_paths.flat_map do |path|
         if File.directory?(path)
           Pathname.glob(File.join(path, RUBY_FILES))
-        elsif File.exist?(path) && File.extname(path) == RUBY_EXTENSION
+        elsif File.exist?(path) && ruby_file?(path)
           Pathname.new(path)
         end
       end.compact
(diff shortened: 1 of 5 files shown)


lib/rubycritic/source_locator.rb
```diff
@@ -7,6 +7,7 @@ module RubyCritic
   class SourceLocator
     RUBY_EXTENSION = '.rb'.freeze
```
jtgrenz, line 8: This will still skip files ending in `.rake` or `.thor`. Maybe it would be better to add a `ruby_extensions` key to the .rubycritic.yml config file and source that value here as an array and give it a good default like `%w(.rb .rake .thor)`. That way anyone wishing to check other file extensions that also don't use a shebang, they're all set, just add it to the config. 

Looks goo other wise :)

etagwerker: @jtgrenz Good point. Thanks for looking into this!

etagwerker (changes requested):
@stufro Thanks for sending this over and sorry for the delay! 

Could you please review @jtgrenz's comment? I believe they raise a valid point and you could be more flexible in terms of extensions.

Other than that this looks good 👍

lib/rubycritic/source_locator.rb
stufro: Sorry for the long delay in getting back to this. I've added a commit to make the ruby extensions configurable. Let me know if there's anything else we could improve.

JuanVqz:
@stufro seems like this PR has conflicts, could you please solve them? Thank you!

stufro:
Solved the conflicts.. seems CI is now failing. Will try to take a look sometime this week.

JuanVqz:
> @JuanVqz I have fixed the build now and rebase the branch. Is there anything else required before this can be merged?

@stufro thank you! @etagwerker could you please review this PR again?

etagwerker (approved):
@stufro Looks good, thanks!

PR #395, opened 2021-06-15
Add logo.svg

AlexVKO:
In case you guys find it useful...(We vectorized it to use as a flat icon)


etagwerker (approved):
@AlexVKO Great, thanks! 👍

PR #399, opened 2021-11-29
Bugfix: Exit with 0 or 1 from Rake Task

### LINKED ISSUE #214: Rake task should fail if minimum score is too low
I would like the rake task to fail if the minimum score is too low.  Is this hard to do?

RyanSnodgrass:
Fixes #214

The rake task always returns a success `0` code even when it fails. I need it to return a `1` when my code doesn't meet the minimum score for CI purposes.

This will break the rake task behavior for the "it's not a bug, it's a feature" people relying on this. I've added a `fail_on_error` attribute to provide backwards compatibility. If they set that to `false` then the rake task reports that the minimum score wasn't met, but still exits with a success.

This matches how [rubocop](https://github.com/rubocop/rubocop/blob/master/lib/rubocop/rake_task.rb#L43) and [reek](https://github.com/troessner/reek/blob/master/lib/reek/rake/task.rb#L108) handle this.

Included is a cucumber spec detailing this change.

Check list:

### DIFF
--- a/.reek.yml
+++ b/.reek.yml
@@ -156,6 +156,7 @@ detectors:
     - RubyCritic::Generator::Html::CodeFile
     - RubyCritic::Generator::Html::Overview
     - RubyCritic::Generator::Html::SmellsIndex
+    - RubyCritic::RakeTask
   ControlParameter:
     exclude:
     - RubyCritic::CommandFactory#self.command_class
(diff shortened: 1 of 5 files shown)


RyanSnodgrass:
Merge me second 😄

nunosilva800 (approved):
Good one 👌

PR #401, opened 2022-03-04
GitHub Actions parses 3.0 as '3', so use quotes

enomotodev:


etagwerker (changes requested):
@enomotodev Thanks for submitting this! I think this is a good idea. 

Would you please do this for all the values under `ruby-version`? I would like us to be consistent with that (not just for 3.0) 

Thanks again!

nunosilva800 (approved):

etagwerker (approved):

PR #405, opened 2022-05-06
Prepare v4.7.0

fbuys:
Check list:
- [x] Add an entry to the [changelog](/CHANGELOG.md)
- [x] [Squash all commits into a single one](/CONTRIBUTING.md)
- [x] Describe your PR, link issues, etc.

This PR closes https://github.com/whitesmith/rubycritic/issues/404.

We updated the changelog and rubycritic/version in preparation for the v4.70 release.

PR #407, opened 2022-05-09
Drop support for ruby 2.4.x

### LINKED ISSUE #406: Drop support for Ruby 2.4.x
A couple of things that would be needed: 

- [ ] Drop ruby 2.4.x from test suite config -- Stop running the test suite with ruby 2.4
- [ ] Add documentation to the README to point users in the right direction so they can figure out if their ruby version is supported
- [ ] Add required_ruby_version to the gemspec file and make it > 2.4

rishijain:
Fixes #406 

Check list:
- [ :heavy_check_mark: ] Add an entry to the [changelog](/CHANGELOG.md)
- [ ] [Squash all commits into a single one](/CONTRIBUTING.md)
- [ :heavy_check_mark: ] Describe your PR, link issues, etc.

### DIFF
--- a/rubycritic.gemspec
+++ b/rubycritic.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |spec|
   spec.summary       = 'RubyCritic is a Ruby code quality reporter'
   spec.homepage      = 'https://github.com/whitesmith/rubycritic'
   spec.license       = 'MIT'
-  spec.required_ruby_version = '>= 2.4.0'
+  spec.required_ruby_version = '> 2.4.0'
 
   spec.files = [
     'CHANGELOG.md',
--- a/rubycritic.gemspec
+++ b/rubycritic.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |spec|
   spec.summary       = 'RubyCritic is a Ruby code quality reporter'
   spec.homepage      = 'https://github.com/whitesmith/rubycritic'
   spec.license       = 'MIT'
-  spec.required_ruby_version = '> 2.4.0'
+  spec.required_ruby_version = '>= 2.5.0'
 
   spec.files = [
     'CHANGELOG.md',
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -29,17 +29,15 @@ end
 
 def with_cloned_fs
   FakeFS do
-    begin
-      FakeFS::FileSystem.clone(PathHelper.project_path)
+    FakeFS::FileSystem.clone(PathHelper.project_path)
 
-      # reek schema is required to init reek
-      FakeFS::FileSystem.clone(PathHelper.reek_schema_path)
+    # reek schema is required to init reek
+    FakeFS::FileSystem.clone(PathHelper.reek_schema_path)
 
-      Dir.chdir(PathHelper.project_path)
-      yield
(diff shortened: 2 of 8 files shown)


rishijain:
Hey @etagwerker, 
instead of auto-running the build, it says this 
`First-time contributors need a maintainer to approve running workflows`

Will you need anything from me to make this happen?

rubycritic.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |spec|
   spec.summary       = 'RubyCritic is a Ruby code quality reporter'
   spec.homepage      = 'https://github.com/whitesmith/rubycritic'
   spec.license       = 'MIT'
-  spec.required_ruby_version = '>= 2.4.0'
+  spec.required_ruby_version = '> 2.4.0'
```
nunosilva800, line 17: Let's make it like so:

```suggestion
  spec.required_ruby_version = '>= 2.5.0'
```

I'm worried that with `'> 2.4.0'` alone, versions like `2.4.1` would be allowed

nunosilva800 (commented):
Thanks @rishijain for your first contribution! I've allowed the tests to go through.
There's only a small consideration here, let me know what you think.

rubycritic.gemspec
rishijain: Yes absolutely. Its a miss from my end. It should be what you suggest. will make a change and push.

rishijain: Hey @nunosilva800 , I have pushed the change. Thanks for the suggestion.

rishijain:
Hey @nunosilva800,
I have tried to make sense of the failures in build https://github.com/whitesmith/rubycritic/actions/runs/2301117538 
but no luck so far.

Can you help me understand what broke this?

Also, I saw similar failures in the last release https://github.com/whitesmith/rubycritic/pull/405. So this suggests that these failures may not be related to any change I did.

What do you think?

test/test_helper.rb
```diff
@@ -29,17 +29,15 @@ def capture_output_streams
...
-      Dir.chdir(PathHelper.project_path)
-      yield
-    ensure
-      FakeFS::FileSystem.clear
-    end
+    Dir.chdir(PathHelper.project_path)
+    yield
+  ensure
+    FakeFS::FileSystem.clear
   end
```
rishijain, line 41: removing the `begin` from the block since th target ruby version in rubocop.yml is set to 2.5 and from ruby 2.5 onwards begin block is seemed redundant as per rubocop rules.
here is that PR when the change was done in rubocop https://github.com/rubocop/rubocop/pull/5175

nunosilva800:
@rishijain I'm still unsure on the exactly why builds for ruby 2.5 are failing. Since Ruby 2.5 is also no longer maintained, I think we can also remove it from the build steps. I'll merge this PR in and raise a new issue. Thanks!

rishijain:
awesome. thanks, @nunosilva800 . 
I will be happy to take that up as well since it would be similar changes. I see that even ruby 2.6 is EOL, why don't we remove both 2.5 and 2.6?

nunosilva800:
2.6 is EOL, but if the CI is still happy with it, I'd say keep it in there. We just not bother with fixing things when the underlying infra / dependencies stop working.

Issue #146, opened 2016-06-17
Missing status label (old/new)


nunosilva800:
![RubyCritic smells index screenshot](http://i.imgur.com/PvkFEe3.png)

Latest version is not generating these "labels" on the rightmost columns.
Not sure when we lost this.

vzamanillo:
I think is related to #115, with that changes the serialized dump will never exists, so there is nothing to compare.

itsmeurbi:
I'd like to address this one

PR #115, opened 2016-01-29
Fix Rubycritic's Analysers running twice on the first run

Onumis:
Fix #94 

This changes `RevisionComparator` so that it does not attempt any comparison if there is nothing to compare to!

PR #416, opened 2022-10-23
[FIX] Restore missing smell status label

itsmeurbi:
This PR will resolve issue https://github.com/whitesmith/rubycritic/issues/146

Check list:
- [ ] [Squash all commits into a single one](/CONTRIBUTING.md)

## Issue description:
Introducen [here](https://github.com/whitesmith/rubycritic/pull/115/files#diff-feea086d0ffc4a6b0de3a916f4aa8bdc2e7c0b423e5deab46857b21e7548d092R18%E2%80%A8)

 `RevisionComparator#statuses=` is skipping `SmellsStatusSetter` if no modules were analyzed before.
As the class name indicates, `SmellsStatusSetter` is responsible for setting the `status` field to smell objects. So because this step is being skipped first time you run rubycritic, `smells` status is always `nil`

## Proposal:
My first thought was to remove the early return and make `analysed_modules_before` an empty array when rubycritics is run for the first time. BUT I believe there was a reason for skipping this step, so I come up with this: 
Consider all new Smell instances as :new(i.e make `status = :new` by default,)

## Screensthots:

Before:
￼
Screen Shot 2022-10-23 at 12 04 54

After:
Screen Shot 2022-10-23 at 12 03 27

￼

Notes:
I’m also fixing an issue I found in `lib/rubycritic/generators/html/templates/smells_index.html.erb`. Basically I removed an extra-no-needed `td` that was creating a new cell

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * [CHANGE] Include files which have a ruby shebang (by [@stufro][])
 * [CHANGE] Fix some typos (by [@ydah][])
 * [CHANGE] Remove wrong Rubocop reference in contributing file (by [@itsmeurbi]: https://github.com/itsmeurbi)
+* [BUGFIX] Restore missing smell status label (by [@itsmeurbi]: https://github.com/itsmeurbi)
 
 # v4.7.0 / 2022-05-06 [(commits)](https://github.com/whitesmith/rubycritic/compare/v4.6.1...v4.7.0)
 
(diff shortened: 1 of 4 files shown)


nefter (approved):

itsmeurbi:
Is there something pending to do here?

etagwerker (approved):
@itsmeurbi Looks good, thank you! 👍

PR #419, opened 2022-11-23
Scope churn calculation to provided paths

Fito:
The churn calculation was hard-coded to `.`
This meant that `rubycritic` would run a `git log` command for the entire repository regardless of the paths specified with running the `rubycritic` command.

Running `git log` for large codebases with many years of commits can take over 30 minutes. This is particularly wasteful in mono-repositories with distinct projects that might be better analyzed separately (large parts of the repository might not even be Ruby code!)

The time to generate a report decreases significantly when scoping churn to the paths provided to the `rubycritic` command.

Other things to note:
- Git log's --follow option can only handle a single path at a time. Therefore we need to run multiple git log commands when multiple path arguments are provided. The results are still merged together.

- The initializer for SourceControlSystem::Git::Churn was already at the maximum number of statements allowed. I extracted `renames` into a memoized method to make room for the new parameter.

- Added tests for SourceControlSystem::Git#churn to ensure SourceControlSystem::Git::Churn is instantiated with the correct options. This also adds a bit of coverage on Config.paths needing to exist.

Check list:

### DIFF
--- a/lib/rubycritic/configuration.rb
+++ b/lib/rubycritic/configuration.rb
@@ -10,7 +10,7 @@ module RubyCritic
                   :feature_branch, :base_branch_score, :feature_branch_score,
                   :base_root_directory, :feature_root_directory,
                   :compare_root_directory, :threshold_score, :base_branch_collection,
-                  :feature_branch_collection, :churn_after, :ruby_extensions
+                  :feature_branch_collection, :churn_after, :ruby_extensions, :paths
 
     def set(options)
       self.mode = options[:mode] || :default
@@ -20,9 +20,14 @@ module RubyCritic
       self.open_with = options[:open_with]
       self.no_browser = options[:no_browser]
       self.threshold_score = options[:threshold_score].to_i
-      self.ruby_extensions = options[:ruby_extensions] || %w[.rb .rake .thor]
+      setup_analysis_targets(options)
       setup_version_control(options)
       setup_formats(options)
+    end
+
+    def setup_analysis_targets(options)
+      self.paths = options[:paths] || ['.']
+      self.ruby_extensions = options[:ruby_extensions] || %w[.rb .rake .thor]
     end
 
     def setup_version_control(options)
--- a/lib/rubycritic/source_control_systems/git.rb
+++ b/lib/rubycritic/source_control_systems/git.rb
@@ -30,7 +30,7 @@ module RubyCritic
       end
 
       def churn
-        @churn ||= Churn.new(churn_after: Config.churn_after)
+        @churn ||= Churn.new(churn_after: Config.churn_after, paths: Config.paths)
       end
 
       def revisions_count(path)
--- a/.reek.yml
+++ b/.reek.yml
@@ -69,6 +69,7 @@ detectors:
     - RubyCritic::Configuration#mode
     - RubyCritic::Configuration#no_browser
     - RubyCritic::Configuration#open_with
+    - RubyCritic::Configuration#paths
     - RubyCritic::Configuration#source_control_system
     - RubyCritic::Configuration#suppress_ratings
     - RubyCritic::Configuration#base_branch
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * [CHORE] Test gem with Ruby 3.2 (by [@etagwerker][])
 * [CHORE] Test gem with Ruby 3.1 (by [@etagwerker][])
+* [CHANGE] Scope churn calculation to provided paths (by [@Fito][])
 * [CHANGE] Drop support for Ruby 2.5.x (by [@cleicar][])
 * [CHANGE] Drop support for Ruby 2.4.x (by [@rishijain][])
 * [FEATURE] Add ruby_extensions configuration option (by [@stufro][])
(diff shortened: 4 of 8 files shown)


lib/rubycritic/source_control_systems/git.rb
```diff
@@ -30,7 +30,7 @@ def self.to_s
       end
 
       def churn
-        @churn ||= Churn.new(churn_after: Config.churn_after)
+        @churn ||= Churn.new(churn_after: Config.churn_after, paths: Config.paths)
```
etagwerker, line 33: @Fito I'm not sure about this. How does this work with the `.rubycritic.yml` configuration file?

etagwerker (commented):
@Fito Thanks for submitting this PR! I'm curious about these changes. How do they work in terms of the existing ways of specifying a path in the command line or in the ruby critic configuration file? It seems that there is a disconnect there. 

Could you please explain a little bit about that?

lib/rubycritic/configuration.rb
```diff
@@ -10,10 +10,11 @@ class Configuration
                   :feature_branch, :base_branch_score, :feature_branch_score,
                   :base_root_directory, :feature_root_directory,
                   :compare_root_directory, :threshold_score, :base_branch_collection,
-                  :feature_branch_collection, :churn_after, :ruby_extensions
+                  :feature_branch_collection, :churn_after, :ruby_extensions, :paths
 
     def set(options)
       self.mode = options[:mode] || :default
+      self.paths = options[:paths] || ['.']
```
etagwerker, line 17: @Fito Maybe this should [default to this](https://github.com/whitesmith/rubycritic/blob/cafe4c92f4a49cede49ab6cb3231b1af651c7466/lib/rubycritic/cli/options/file.rb#L58-L60)?

lib/rubycritic/source_control_systems/git.rb
Fito: @etagwerker This change works both for the `paths` options defined in `.rubycritic.yml` and for paths provided via the command line.

The `RubyCritic::Cli::Options#to_h` method merges all options provided via the command line and via the `.rubycritic.yml` configuration file. [here](https://github.com/whitesmith/rubycritic/blob/main/lib/rubycritic/cli/options.rb#L23)

`RubyCritic::Cli::Application` instantiates an `RubyCritic::Cli::Options` object and calls `to_h` on it before passing it into `RubyCritic::CommandFactory.create` to execute the reporter [here](https://github.com/whitesmith/rubycritic/blob/main/lib/rubycritic/cli/application.rb#L21).

The `RubyCritic::CommandFactory` sets the options hash on the global config object via `RubyCritic::Config.set(options)` [here](https://github.com/whitesmith/rubycritic/blob/main/lib/rubycritic/command_factory.rb#L10).

This means `Config.paths` will contain the paths provided to RubyCritic via either CLI or Config File options.

lib/rubycritic/configuration.rb
Fito: The `path` (singular) option currently refers to the path where the reports will get written. The `paths` (plural) option refers to the paths to be analyzed by RubyCritic. There might be an opportunity to rename either of those concepts for clarity, but it feels outside the scope of this change. The names were already in the codebase.

Fito:
@etagwerker Thank you for reviewing the changes! I replied to your comments on specific changes. The high-level answer is that this change respects and works with existing ways of specifying paths to be analyzed. My change is only concerned with the `paths` option which refers to paths in the given codebase that should be analyzed.
The changes are _not_ concerned with the singular `path` option which refers to the path where RubyCritic reports are written.

Let me know if you have any other questions!

Fito:
I pushed a small change to address a couple of Rubocop violations my commit added.

Fito:
The remaining CI failures don't seem related to my change. Is it normal for some checks to fail?

Fito:
@etagwerker Do you have any more feedback on this?

etagwerker (changes requested):
@Fito I think the changes look good. The last thing I would ask is that you document this new parameter somewhere in the README. 

Considering this PR adds a new parameter that is used only with the Git SCM, I think it should be clear that this parameter will not have any effect if you are using another SCM (e.g. Mercurial)

Other than that it looks good 👍🏼

etagwerker:
@Fito I just resolved the conflict in this PR but now it seems to be failing in a couple scenarios that are related to the new parameter. Could you please take a look? Thanks!

Fito:
@etagwerker Thanks for resolving the conflicts, and for the suggestion about the README. I will take a look at the failing scenarios.

Fito:
@etagwerker I fixed the failing test and added a bit more about the "paths" option to the README.
I want to clarify though that this is NOT at new parameter. The "paths" parameter was already part of the gem before my change, you can see it documented in the Readme on the main branch -- However, it was getting ignored by the churn analyzer for Git. This commit makes the analyzer use the option when provided. Does that make sense?

Thanks again for the review and for fixing the merge conflicts! Please let me know if you have any other questions.

etagwerker (approved):
@Fito Thank you! 👍🏼

Fito:
Thanks for reviewing this @etagwerker !

Issue #346, opened 2020-01-30
Shell error with rake task


lloydwatkin:
```
!!! Running `rubycritic` rake command
!!! Inspecting app/errors app/jobs app/serializers app/tasks app/assets app/exceptions app/mailers app/services app/uploaders app/controllers app/helpers app/models app/states app/views with options --mode-ci --format json --no-browser
Warning: you are leaving 5 commits behind, not connected to
any of your branches:
  045b88d Merge d43ecf263cf6f45b7f1981e07cd489b9e7d23a52 into dec6daea97a124a57a5c599768e2fdca7ab99011
  d43ecf2 Fix up tests
  171d4a4 Update list of what is being observed
  ebbc596 Missed the addition of the observer
  a5353a5 Move mixpanel/mailchimp updates to on-demand observer setup
If you want to keep them by creating a new branch, this may be a good time
to do so with:
 git branch <new-branch-name> 045b88d
Switched to branch 'master'
running flay smells
.................................................
running flog smells
........................................................................................................................................................................................................................................................................................................................................................................................
running reek smells
........................................................................................................................................................................................................................................................................................................................................................................................
running complexity
........................................................................................................................................................................................................................................................................................................................................................................................
running attributes
........................................................................................................................................................................................................................................................................................................................................................................................
running churn
........................................................................................................................................................................................................................................................................................................................................................................................
running simple_cov
........................................................................................................................................................................................................................................................................................................................................................................................
sh: 1: Syntax error: "(" unexpected
running flay smells
.................................................
running flog smells
........................................................................................................................................................................................................................................................................................................................................................................................
running reek smells
........................................................................................................................................................................................................................................................................................................................................................................................
running complexity
........................................................................................................................................................................................................................................................................................................................................................................................
running attributes
........................................................................................................................................................................................................................................................................................................................................................................................
running churn
........................................................................................................................................................................................................................................................................................................................................................................................
running simple_cov
........................................................................................................................................................................................................................................................................................................................................................................................
sh: 1: Syntax error: "(" unexpected
Score (87.54) is below the minimum 95
The command "bundle exec rspec && bundle leak check --update && bundle audit check --update && git reset --hard && bundle exec rake rubycritic" exited with 0.

etagwerker:
@lloydwatkin Weird. What version of `rubycritic` are you using?

lloydwatkin:
Latest 2.4.3

etagwerker:
[...]

PR #424, opened 2023-01-11
Support a branch in detached HEAD state

h-r-k-matsumoto:
This PR would resolve issue #346

The code below has a problem if check out a branch in the state of `detached HEAD`.
https://github.com/whitesmith/rubycritic/blob/main/lib/rubycritic/source_control_systems/git.rb#L82-L85

Even in that case, it can be run successfully with this change.

Check list:


etagwerker (approved):
@h-r-k-matsumoto I like this idea. Thank you! ❤️

PR #430, opened 2023-01-25
Add Config.command to simplify the command_factory file

JuanVqz:
Check list:

### DIFF
--- a/lib/rubycritic/command_factory.rb
+++ b/lib/rubycritic/command_factory.rb
@@ -4,22 +4,14 @@ require 'rubycritic/configuration'
 
 module RubyCritic
   class CommandFactory
-    COMMAND_CLASS_MODES = %i[version help ci compare default].freeze
-
     def self.create(options = {})
       Config.set(options)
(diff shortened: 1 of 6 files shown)


lib/rubycritic/command_factory.rb
```diff
@@ -6,19 +6,28 @@ module RubyCritic
...
-      mode = mode.to_s.split('_').first.to_sym
-      if COMMAND_CLASS_MODES.include? mode
-        require "rubycritic/commands/#{mode}"
-        Command.const_get(mode.capitalize)
-      else
-        require 'rubycritic/commands/default'
-        Command::Default
+      def mode
+        mode = Config.mode
+        mode.to_s.split('_').first.to_sym
```
JuanVqz, line 30: I wonder if we can add a new Config attribute called `command`, to use it like so `Config.command` that way we can delegate all this parse to a class and save it there.

JuanVqz: In the end, I went ahead and changed it, let's see what do you think about it.

PR #440, opened 2023-03-22
Update jruby CI configuration

etagwerker:
Check list:
- [ ] Add an entry to the [changelog](https://github.com/whitesmith/rubycritic/blob/main/CHANGELOG.md)

I'm pretty sure RubyCritic no longer works with jruby-9.2 due to recent required ruby version changes to the gem's configuration. So this is an attempt to update jruby's configuration (experimental) in CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,9 +37,9 @@ jobs:
         include:
           - ruby-version: 'ruby-head'
             experimental: true
-          - ruby-version: 'jruby-9.2'
-            experimental: true
           - ruby-version: 'jruby-9.3'
+            experimental: true
+          - ruby-version: 'jruby-9.4'
             experimental: true
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


h-r-k-matsumoto:
Hi @etagwerker ,
Sorry if I might be rude.
You faced the following problem.
https://github.com/jruby/jruby/issues/7735

This is a problem of JRuby or fakefs, but we can resolve it with following patch.
https://github.com/h-r-k-matsumoto/rubycritic/commit/b1d6035eb58b75ae432fb1964122d99d5dcf9556
https://github.com/h-r-k-matsumoto/rubycritic/actions/runs/4520028160

etagwerker:
@h-r-k-matsumoto Thanks for that patch, it totally worked! 

Do you think that should be a patch submitted to `fakefs` itself?

h-r-k-matsumoto:
@etagwerker Thanks for your reply.

> Do you think that should be a patch submitted to `fakefs` itself?

It's hard to answer...
But I'm afraid that my answer is No.
It's because that the root cause of this problem is `JRuby` so users should upgrade `JRuby`.
And more, I don't know `fakefs` support `JRuby` or not.

But using the version of `fakefs` with `rubycritic` is a bit old in order to supporting the old Ruby version.
In that case, we have to address on the issue individually.

This is my opinion.

etagwerker:
@h-r-k-matsumoto got it, thank you!

PR #458, opened 2023-05-16
Remove Ruby 2.7 support

faisal:
This is for the move to Ruby 3, for when we're ready to do that.

- Remove Ruby 2.7 support.
- Set Rubocop to target Ruby 3. Exclude lib/rubycritic/configuration.rb from Metrics/AbcSize cop.

Check list:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,7 +29,6 @@ jobs:
       fail-fast: false
       matrix:
         ruby-version:
-          - '2.7'
           - '3.0'
           - '3.1'
           - '3.2'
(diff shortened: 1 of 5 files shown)


faisal:
When doe sit makeaWhen does it make sense to drop 2.7 support?

nunosilva800:
Not sure. Just because ruby 2.7 reached end-of-life, it does not mean we need to drop it from CI.
I think that until CI breaks or any of the direct dependencies require ruby 3.0, we can keep testing against ruby 2.7 and listing it as officially supported.

etagwerker:
I agree with @nunosilva800 on this. Maybe we can keep support for now.

PR #463, opened 2023-09-05
WIP: Address test warning by declaring cucumber_opts using %w[] instead of…

faisal:
Check list:

This is a minor change in the Rakefile, clearing a test warning (see [465](https://github.com/whitesmith/rubycritic/issues/465)) by declaring `cucumber_opts` using `%w[]` syntax instead of a raw string.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # main [(unreleased)](https://github.com/whitesmith/rubycritic/compare/v4.8.1...main)
 
+* [CHANGE] Fix test warning related to `cucumber_opts` declaration (by [@faisal][])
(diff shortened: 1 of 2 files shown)


faisal:
https://github.com/minitest/minitest/issues/963

https://my.diffend.io/gems/minitest/5.18.1/5.19.0#d2h-230603

Looking to fix this in a way that's compatible across releases. Failing that, I'll lock the minitest dependency to the earlier version so we have the minimal change here.

faisal:
https://github.com/whitesmith/rubycritic/pull/464 fixes the issue that caused test break here. If we land that fix then we should undo the minitest version restriction in this PR before merging.

nunosilva800 (approved):

PR #464, opened 2023-09-05
BUGFIX: use Minitest rather than deprecated MiniTest constant

faisal:
Minitest 5.19.0 only loads the legacy compatibility layer if you have the MT_COMPAT environment variable set -- https://github.com/minitest/minitest/commit/a2c6c18570f6f0a1bf6af70fe3b6d9599a13fdd6. This causes use of the MiniTest module name to fail, as well as the bundled older mocha. Fix our code to use the correct Minitest name, and update to a fixed mocha.

Check list:


faisal:
@nunosilva800, @etagwerker can you take a look please?

nunosilva800 (approved):
Nice!

Issue #465, opened 2023-09-11
Clean test warnings


faisal:
```
$ bundle exec rake
WARNING: consider using an array rather than a space-delimited string with cucumber_opts to avoid undesired behavior.
/Users/faisal/Documents/Versioned/rubycritic/test/fakefs_helper.rb:8: warning: method redefined; discarding old flock
/Users/faisal/Documents/Versioned/rubycritic/.bundle/ruby/3.2.0/gems/fakefs-2.4.0/lib/fakefs/file.rb:586: warning: previous definition of flock was here
/Users/faisal/Documents/Versioned/rubycritic/test/lib/rubycritic/commands/compare_test.rb:12: warning: method redefined; discarding old current_branch
/Users/faisal/Documents/Versioned/rubycritic/lib/rubycritic/source_control_systems/git.rb:82: warning: previous definition of current_branch was here
```

...

The first looks trivial. The remainder look harder unless @etagwerker thinks there are now better ways to test the behavior for which he put them in.

fbuys:
Hey! I opened a PR for the remaining warnings:
https://github.com/whitesmith/rubycritic/pull/466

fabiosoaresv:
I think this issue can be close ?

faisal:
> I think this issue can be close ?

Tested and confirmed.

PR #466, opened 2023-10-02
ISSUE-456: Disable VERBOSE warnings in test stubs

fbuys:
To prevent warning during tests while we deliberately override certain methods.

Warnings include:
rubycritic/test/fakefs_helper.rb:8: warning: method redefined; discarding old flock rubycritic/.bundle/ruby/3.2.0/gems/fakefs-2.4.0/lib/fakefs/file.rb:586: warning: previous definition of flock was here rubycritic/test/lib/rubycritic/commands/compare_test.rb:12: warning: method redefined; discarding old current_branch rubycritic/lib/rubycritic/source_control_systems/git.rb:82: warning: previous definition of current_branch was here

See: https://github.com/whitesmith/rubycritic/issues/465

An alternative would be to turn warning completely off during test, but I believe warnings to play an important role.
In this case turning them off temporarily seems like a good idea because we are overriding methods specifically for tests.

Check list:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * [CHANGE] Fix test warning related to `cucumber_opts` declaration (by [@faisal][])
 * [BUGFIX] Stop using long-deprecated MiniTest module name, removed in 5.19.0 (by [@faisal][])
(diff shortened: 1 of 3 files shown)


etagwerker (approved):
@fbuys Looks good, thank you!

Issue #461, opened 2023-07-10
Running in subdirectory returns `churn` of `0`


geetfun:
When running rubycritic in a subfolder, the Churn returns `0`.

Example of a mono-repo where a subfolder is a Rails app, and other folders hold other components (eg. React app, docs, etc):

main-app
|-rails_app
|-docs
|-frontend

```
cd rails_app
rubycritic # Returns churn of 0
```

Running from the root directory works.

Steps to reproduce:
1. Git clone sample app (https://github.com/geetfun/rails-in-subdir)
2. Run rubycritic from the root folder --> No problem
3. Run rubycritic from the Rails directory subfolder --> Churn shows 0

From the root directory:

<img width="1497" alt="Screenshot 2023-07-10 at 7 35 13 PM" src="https://github.com/whitesmith/rubycritic/assets/16937/275b5a0c-b42f-46bd-aace-e45ef5535e70">

From the subfolder:
<img width="1496" alt="Screenshot 2023-07-10 at 7 35 02 PM" src="https://github.com/whitesmith/rubycritic/assets/16937/40a71351-8d1f-4ec5-b042-be5aeacdabe5">

Expectation: Both churn values should be the same regardless of where the rubycritic command is run

Side thought: This may just be a limitation of how the app calls git within a subfolder, but it's a gotcha when working with monorepos.

rishijain:
@etagwerker  Can we close this issue ?

etagwerker:
Fixed in v4.9.0 via #468 ("churn score was always 0 when rubycritic was executed from a sub-directory"). Closing.

PR #468, opened 2023-10-11
Issue 461 - Fix for churn score when accessed from a sub-directory

rishijain:
This is a fix for https://github.com/whitesmith/rubycritic/issues/461.

From my research, I understood that there are 2 ways to fix this. 
1) When the score is already set for the paths, so when you are trying to fetch the churn score of a file, adjust the path of the file to match the file name that is being stored in the @stats variable.

2) Adjust the filename when the score is being set, and then fetch the score of the file without any manipulation.

I decided to go with 2nd approach as it seems more cleaner way of solving this.

Check list:


nunosilva800 (approved):
LGTM

etagwerker (approved):

Issue #469, opened 2023-10-16
Links pointing to pages that do not exist when using the Compare branch option.


rishijain:
When you run the command `rubycritic -b test_branch` with `main` branch.
So the page which loads upon running this command have 3 links at top right corner.  The links are "main",  "test_branch", and "build" in my case.

Now lets say you have a file called "app/models/contact.rb" in the application and there is no difference in the code of that file for the 2 branches "main" and "test_branch".

* You click on "main" tab at the top right corner.
* Click on the `Code` tab from the left navigation
* Click on the file `app/models/contact.rb`
* Click on the "build" tab at the top right corner

This opens a page which does not exist as there was no difference in the code of the 2 branches for this file, it does not really exist for the "build" tab. 
<img width="609" alt="Screenshot 2023-10-16 at 8 06 32 PM" src="https://github.com/whitesmith/rubycritic/assets/946527/18c9cd0a-e7a6-4cc6-a61b-f9c0ab402346">

I think the behaviour should be that whenever the user clicks on the "build" tab for a file which does not exist, we should instead open the "overview.html" file.

rishijain:
Closing this since the pr was merged.

PR #470, opened 2023-10-16
Issue 469 - Fallback to overview.html when pages do not exist for compare option

rishijain:
Check list:
- [ x ] Add an entry to the [changelog](https://github.com/whitesmith/rubycritic/blob/main/CHANGELOG.md)
- [ x ] [Squash all commits into a single one](https://github.com/whitesmith/rubycritic/blob/main/CONTRIBUTING.md)
- [ x ] Describe your PR, link issues, etc.

Description: This fixes issue #469 .

### DIFF
--- a/lib/rubycritic/generators/html/view_helpers.rb
+++ b/lib/rubycritic/generators/html/view_helpers.rb
@@ -24,7 +24,10 @@ module RubyCritic
     end
 
     def code_index_path(root_directory, file_name)
-      file_path("#{File.expand_path(root_directory)}/#{file_name}")
+      root_directory_path = File.expand_path(root_directory)
+      index_path = "#{root_directory_path}/#{file_name}"
+      index_path = "#{root_directory_path}/overview.html" unless File.exist?(index_path)
+      file_path(index_path)
     end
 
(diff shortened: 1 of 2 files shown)


lib/rubycritic/generators/html/view_helpers.rb
```diff
@@ -24,7 +24,10 @@ def smell_location_path(location)
     end
 
     def code_index_path(root_directory, file_name)
-      file_path("#{File.expand_path(root_directory)}/#{file_name}")
+      root_directory_path = File.expand_path(root_directory)
+      file_path = "#{root_directory_path}/#{file_name}"
+      file_path = "#{root_directory_path}/overview.html" unless File.exist?(file_path)
+      file_path(file_path)
```
nunosilva800, line 30: This is weird. What about naming the local var `index_path` or just `path` so not confuse it with the method name `file_path`?

rishijain: @nunosilva800  oh yes my apologies. How did I not see how bad that naming was. I have used index_path.

Issue #423, opened 2022-12-06
Coverage index page of "Build" shows 0 coverage if compare view 


exoego:
If compare view (`--branch BRANCH` or `--mode-CI BRANCH` is used), a coverage index page shows the names of changed files, with `0%` coverage.

<img width="255" alt="image" src="https://user-images.githubusercontent.com/127635/205794002-10a220e2-2bf9-49d4-9b92-f817946ca4ed.png">
If I switch to the `main` or `rubycritic-setup` branch from the above menu, their coverage index pages shows non-zero coverages.

"Build" `compare/simple_cov_index.html` | target branch `compare/main/simple_cov_index.html`| source branch `compare/rubycritic-setup/simple_cov_index.html`
---|---|---
![image](https://user-images.githubusercontent.com/127635/205793253-232229aa-f412-495e-bcd9-7c6a3c355568.png) | ![image](https://user-images.githubusercontent.com/127635/205793464-40c26b1a-1336-4a33-8ffc-cebcf5b0b0e2.png) | ![image](https://user-images.githubusercontent.com/127635/205793826-dc0322b1-0145-4979-b551-ea779cdca9b7.png)

 I expected the "Build" coverage index shows the coverages of the changed files.

PR #474, opened 2023-10-19
add coverage data to compare view

rishijain:
Check list:

This fixes issue: https://github.com/whitesmith/rubycritic/issues/423
I noticed that we are not passing any coverage data and the default value of it being 0, it shows 0% against all the files for compare build view.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -85,3 +85,7 @@ Lint/StructNewOverride:
 Metrics/AbcSize:
   Exclude:
     - 'lib/rubycritic/configuration.rb'
+
+Metrics/MethodLength:
+  Exclude:
+    - 'lib/rubycritic/core/analysed_modules_collection.rb'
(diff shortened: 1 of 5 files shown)


santhosh-fit:
@nunosilva800 could you review and merge this please? I am starting to use rubycritic and showing the correct coverage info on build/compare would be helpful.

etagwerker (commented):
@rishijain How does this work? I don't understand what's the goal of this PR.

I think it's a tricky problem to solve: To show the difference in coverage between two branches.

The main issue is this: People don't usually check in their coverage directory (so the coverage artifacts are going to be the same across two branches) which makes it very tricky to show a comparison of coverage data.

In order to show a useful comparison in coverage data, then we would have to rely on the user running the test suite in one branch, storing the coverage resultset artifact, then another, then store the new artifact, then run rubycritic and then compare this.

I'm inclined to not support this in the compare view for now... Thoughts? 

PS. Sorry for the delay. :)

Issue #413, opened 2022-08-03
Support Wildcards in Configured Paths


SingleShot:
We are using the [packwerk](https://github.com/Shopify/packwerk) gem which helps with partitioning a Rails app into domain-specific packages. A common way to do this is create a `packs` directory, a sub-directory for each domain, and within each of those, the typical Rails `app`, `controllers`, `models`, `specs`, etc. directories. Due to this, we have to configure Rubycritic like this:
```
paths:
  - app
  - packs/accounts/app
  - ... many more ...
  - packs/users/app
  - packs/widgets/app
```

Something like this would be very nice so we don't have to remember to update the file whenever we add a new package:
```
paths:
  - app
  - packs/*/app
```

Note that if we were to do something like this, Rubycritic picks up all the specs - which we don't want:
```
paths:
  - app
  - packs
```

Thanks!

etagwerker:
Wildcard patterns for `paths` in `rubycritic.yml` shipped in v4.9.1 via #476. Closing.

PR #476, opened 2023-10-25
[Issue - 413 ] Support for wildcard in paths options

rishijain:
Check list:

### Description:
This PR adds support for passing **(wildcard) paths to the .rubycritic.yml file. So now someone can do something like:

```
paths:
  - app
  - packs/*/app
```

This was requested [here](https://github.com/whitesmith/rubycritic/issues/413).

The way I have implemented is to basically find all the valid directories that the ** expands to, so for example:
if user passes `packs/*/app` and in the app, `packs` has 3 sub-directories in it but only 2 of them has `app` inside those sub-directories, I am resetting the options[:paths] to `['app', 'packs/pack1/app', 'packs/pack2/app' ]`.

### DIFF
--- a/.reek.yml
+++ b/.reek.yml
@@ -92,9 +92,11 @@ detectors:
   DuplicateMethodCall:
     exclude:
     - RubyCritic::Analyser::Churn#run
+    - RubyCritic::Configuration#set
     - Parser::AST::Node#module_name
   TooManyStatements:
     exclude:
+    - RubyCritic::Configuration#find_directories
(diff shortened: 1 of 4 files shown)


etagwerker (changes requested):
@rishijain This looks good, thanks! 

Could you please add a note to the README to make sure people understand there is now support for wildcards in paths? Also, could you please add one or two test cases to test this new behavior?

Thank you! ❤️

etagwerker (approved):

PR #492, opened 2024-06-07
Drop support for Ruby 3.0, and adopt Rubycritic changes for Ruby 3.1.

faisal:
While we have not yet dropped support for Ruby 2.7 (EOL'd over a year ago), Ruby 3.0 EOL'd a couple months ago and this PR does the work to move forward from there.

- Drop support for Ruby 3.0.
- Update Rubycritic config to target Ruby 3.1. Apply all safe autocorrections except for Style/HashSyntax.
- Set Rubycritic's Style/HashSyntax EnforcedShorthandSyntax to 'never', which keeps the old style in place (and consistent) until we decide to convert.
- Add Ruby 3.3 to the test matrix and Drop Ruby 3.0.
- Bump minitest, mocha, rubocop dependencies

Check list:


faisal:
@etagwerker @nunosilva800 Regarding Rubocop's preference `EnforcedShorthandSyntax`, I set that to `either` so as to minimize the change in this PR. Another option would be to adopt that default setting, and let `rubocop -a` do the safe right thing. What do you think makes sense here?

faisal:
The failure in 3.4 seems to be an issue with running cucumber on ruby 3.4-dev, and unrelated to this project. I can reproduce it elsewhere, even on null projects.

faisal:
https://github.com/cucumber/cucumber-ruby/issues/1769

etagwerker (approved):
@faisal Looks good to me, thank you! 🙏🏼

PR #500, opened 2024-12-24
Bump dependencies

faisal:
- Bump dependencies to the latest available that pass with all currently-supported Rubies.
- Add gems that have been removed from the standard library in Ruby HEAD.

Note that the CHANGELOG will probably need some reworking before a release. Since we currently have lines for dropping both Ruby 2.7 and 3.0, and lines for dropping two overlapping sets of dependencies.

Check list:


faisal:
Observations on that Cucumber failure:

1. It's the same "this will be fixed in Cucumber 10" bug that was holding up other recent PRs.
2. The tests passed locally on 3.4 RC1

Unless somebody wants these changes now, I plan to wait for Cucumber 10 to land and check with that.

faisal:
I suspect we'll want to hold this open until Cucumber 10 lands.

faisal:
I can reproduce that jruby failure locally. Investigating ...

faisal:
This latest push picks up a newer Cucumber, which includes a limited fix for the keywords issue that is causing things to fail on 3.4. I also need to update the CI to include 3.4, and then investigate this new failure on HEAD.

faisal:
This waits on https://github.com/whitesmith/rubycritic/pull/502.

etagwerker (approved):
@faisal Looks good to me, thank you!

PR #502, opened 2025-01-12
development_dependency the gems Ruby 3.5 drops from the standard library.

faisal:
Ruby 3.5 dropped ostruct and irb from the standard library, causing PRs to fail CI when running on HEAD. This PR adds them back as development dependencies so CI passes again.

Check list:


faisal:
I'm going to combine this one with https://github.com/whitesmith/rubycritic/pull/500, because I don't believe having two "we should update the gems" PRs will make sense after the fact.

PR #508, opened 2025-03-06
Fixes/jruby ruby head

etagwerker:
Hi there,

I'd like to mark the `ruby-head` job as "green" even if it fails. It's adding unnecessary noise when it's red.

Check list:
- [ ] Add an entry to the [changelog](https://github.com/whitesmith/rubycritic/blob/main/CHANGELOG.md)
- [ ] [Squash all commits into a single one](https://github.com/whitesmith/rubycritic/blob/main/CONTRIBUTING.md)


rubycritic.gemspec
```diff
@@ -46,6 +51,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'bundler', '>= 2.0.0'
   if RUBY_PLATFORM == 'java'
     spec.add_development_dependency 'pry-debugger-jruby'
+    spec.add_development_dependency 'jar-dependencies', '~> 0.4.1'
```
faisal, line 49: Given the comment about jar-dependencies in https://www.jruby.org/2025/02/11/jruby-9-4-12-0.html, "jar-dependencies upgraded to 0.5.4 to fix an issue parsing Maven output on Java versions 9 and higher," should we lock jar-dependencies to `-> 0.5.4`, or even remove the version restriction so as to pick up whatever jruby wants to install?

faisal:
This seems worth taking.

faisal:
https://github.com/cucumber/cucumber-ruby-core/pull/292

etagwerker:
@faisal Yes, I'd still like to have this merged into `main`

I know it's a bit of OCD on my side, but I'd like the checks to be green for pull requests where only `ruby-head` fails. 

Screenshot 2025-04-07 at 9 38 17 AM

I thought this change was going to be enough: 

```
  Features:
    needs: pre_job # skip duplicates
    if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
    runs-on: ubuntu-latest
    continue-on-error: ${{ matrix.experimental }}
    strategy:
      fail-fast: false
      matrix:
        ruby-version:
          - '3.1'
          - '3.2'
          - '3.3'
          - '3.4'
        experimental: [false]
        include:
          - ruby-version: 'ruby-head'
            experimental: true
            continue-on-error: true
          - ruby-version: 'jruby-9.4'
            experimental: true
    steps:
      - uses: actions/checkout@v2
      - name: Set up Ruby ${{ matrix.ruby-version }}
        uses: ruby/setup-ruby@v1
        with:
          ruby-version: ${{ matrix.ruby-version }}
          bundler-cache: true
          cache-version: 1
      - name: Run Cucumber
        run: bundle exec cucumber features --format progress --color
```

But it is still not doing what I want. 

Any ideas?

faisal:
[...]
> But it is still not doing what I want.
> 
> Any ideas?

In cursory testing, I haven't been able to find a way to do this without changing runners. See https://github.com/actions/runner/issues/2347.

faisal:
It appears the check mark is always red if anything failed, even if the thing that failed is not required. Beyond OCD, I find this user-hostile because there's no quick way to know which PRs are ready for merging if any tasks aren't required to succeed.

As to how to address this. We could potentially have the experimental items always "succeed" but emit useful logging if they fail, then keep an eye on those logs in the experimental tasks.

Note that this is independent of what failures block a merge. If you have a branch rule that blocks merges you can specify which specific jobs must pass. We could move the experimental tasks out to "experimental" jobs and only require the "supported" jobs pass before merging. I think this factoring is worth doing in either case, and will try to take a swing at it when I get a moment.

faisal:
Something like https://github.com/whitesmith/rubycritic/pull/517 sets up for having the branch rule block merges for the `-supported` jobs and `MarkdownLint` but not the `-experimental` jobs.

@etagwerker what do you think?

Also if you want to keep considering this problem then is it worth putting the gemspec changes into a focussed PR?

faisal:
As an aside, https://github.com/faisal/rubycritic/tree/test_cucumber_should_work squashes the current Ruby 3.5 failure, at least. I think it's a bit 🤡 to check in, but it would get us back to green if that's the goal.

JuanVqz:
Hey @faisal, thanks for the effort you have been putting in this PR. if you continue on it, I can help you get it to the final line; just ping me

faisal:
I believe the work of this PR is now entirely complete:

https://github.com/whitesmith/rubycritic/pull/534 makes experimental branches show green (with annotations) even if they fail

https://github.com/whitesmith/rubycritic/pull/554 introduces @etagwerker's improved Java platform deliniation.

PR #511, opened 2025-03-24
Loosen Rubocop dependency to pick up any new versions < 2.0

faisal:
This change unbinds the version for the Rubocop dependency so it will take any new version prior to 2.0. Rubocop is stable but keeps iterating, so I believe it's worth picking up any new fixes. Since it's a development dependency, doing so should not have a significant risk to the Rubycritic gem itself.

Check list:


faisal:
n.b. jruby magically cooperating here

etagwerker (approved):
Looks good to me. Thanks!

PR #518, opened 2025-04-10
monkeypatch cucumber-core to handle Ruby 3.5 method signatures

faisal:
We get breaks on Ruby 3.5 because the method signature expected for Cucumber::Core::Test::Location#from_source_location changed. This PR monkeypatches the module using the technique being considered in https://github.com/cucumber/cucumber-ruby-core/pull/292, and guards the patch so it only applies on Ruby 3.5 and with Cucumber 13.0.x.

It's either a proof of concept, or something we might tighten further (for example, `RUBY_PATCHLEVEL == -1`) and actually put in the codebase until the Cucumber project fixes things.

Check list:
- [ ] Add an entry to the [changelog](https://github.com/whitesmith/rubycritic/blob/main/CHANGELOG.md) - Not yet. Whether to do this and what to say is going to depend on what we do with this PR and why.


faisal:
@etagwerker This is a hack but do you think it's worth landing it temporarily so as to clean up the CI results on the PR queue?

faisal:
> @etagwerker This is a hack but do you think it's worth landing it temporarily so as to clean up the CI results on the PR queue?

If the answer is yes then I'll also open a ticket (or PR?) to back it out later.

faisal:
Based on https://github.com/cucumber/cucumber-ruby-core/pull/292#issuecomment-2887430046 it sounds like we're in for at least another couple months of HEAD breaking (absent this patch or some other workaround).

FWIW I have this patch in my local copy and have not noticed any regressions.

faisal:
```
$ git diff
diff --git a/rubycritic.gemspec b/rubycritic.gemspec
index 23e53a9..b2cfc5e 100644
--- a/rubycritic.gemspec
+++ b/rubycritic.gemspec
@@ -49,7 +49,7 @@ Gem::Specification.new do |spec|
   else
     spec.add_development_dependency 'byebug', '~> 11.0', '>= 10.0'
   end
-  spec.add_development_dependency 'cucumber', '~> 9.2.1', '!= 9.0.0'
+  spec.add_development_dependency 'cucumber', '~> 10.0.0'
   spec.add_development_dependency 'diff-lcs', '~> 1.3'
   spec.add_development_dependency 'fakefs', '~> 2.6.0'
   spec.add_development_dependency 'irb'
$ bundle update
Fetching gem metadata from https://rubygems.org/.......
Resolving dependencies...
Could not find compatible versions

Because aruba >= 2.2.0 depends on cucumber >= 8.0, < 10.0
  and Gemfile depends on aruba ~> 2.3.0,
  cucumber >= 8.0, < 10.0 is required.
So, because Gemfile depends on cucumber ~> 10.0.0,
  version solving has failed.
```

faisal:
https://github.com/cucumber/aruba/issues/949

Issue #513, opened 2025-03-25
Support for JRuby 10


faisal:
JRuby 10 is coming, with Ruby 3.4 syntax: <https://github.com/jruby/jruby/milestone/95>. Presumably we'll want to support it.

faisal:
https://github.com/whitesmith/rubycritic/pull/521

faisal:
Candidate implementation in https://github.com/whitesmith/rubycritic/pull/514.

etagwerker:
JRuby 10 support landed in v5.0.0 via #554, and CI now runs the full matrix on jruby-10.0. Closing.

faisal:
While we do have JRuby 10 "in" Rubycritic, the CI tends to fail (which shows up in annotations rather than a red result). Can we rightly call that running?

I wonder if we should take JRuby out of the test matrix and return it only once we have it completely working.

PR #512, opened 2025-03-25
Bump byebug and reek dependencies.

faisal:
This picks up newer dependencies for byebug (which adds support for more Ruby versions, plus some fixes and improvements -- <https://github.com/deivid-rodriguez/byebug/blob/master/CHANGELOG.md>) and reek (of possibly lower value -- <https://github.com/troessner/reek/blob/master/CHANGELOG.md>).

Check list:
- [X] Add an entry to the [changelog](https://github.com/whitesmith/rubycritic/blob/main/CHANGELOG.md)
- [X] [Squash all commits into a single one](https://github.com/whitesmith/rubycritic/blob/main/CONTRIBUTING.md)
- [X] Describe your PR, link issues, etc.

PR #504, opened 2025-01-23
Fakefs 3

faisal:
Pick up the FakeFS 3.0 dependency, which should cut down on warnings about frozen strings.

Check list:
- [x] Add an entry to the [changelog](https://github.com/whitesmith/rubycritic/blob/main/CHANGELOG.md)
- [X] [Squash all commits into a single one](https://github.com/whitesmith/rubycritic/blob/main/CONTRIBUTING.md)
- [X] Describe your PR, link issues, etc.

PR #524, opened 2025-06-01
Bump aruba, byebug, cucumber, fakefs, rake, reek dependencies

faisal:
This PR bumps the byebug, fakefs, rake, and reek dependencies. It is effectively a combination of https://github.com/whitesmith/rubycritic/pull/512 and https://github.com/whitesmith/rubycritic/pull/504, plus bumping the rake dependency. I combined them all because it feels like updating them all in individual PRs at this point is just makework. I'll closing out those first to PRs but can update them if you think it does make sense to land things seperately.

N.b. I do not expect CI will pass on Java 9 (still) or 10 (yet – see https://github.com/whitesmith/rubycritic/issues/513) although I don't believe that's a regression in either case. Further, it still won't pass on HEAD unless we do something like https://github.com/whitesmith/rubycritic/pull/518.

Check list:


faisal:
This now depends on the new Cucumber 10. CI still fails on HEAD (worked around in https://github.com/whitesmith/rubycritic/pull/518) on JRuby (tracking work in https://github.com/whitesmith/rubycritic/issues/513), and due to new Rubocop cops (addressed in https://github.com/whitesmith/rubycritic/pull/525).

Issue #537, opened 2025-10-23
[Bug] Excluded files in .reek.yml is not respected


fbuys:
### Current behavior
- I have a .reek.yml file that excludes certain paths
- When I run rubycritic
  - it analyses files with reek that are meant to be excluded
  - it adds smells for files that should be excluded from reek analysis

### Expected behavior
- I have a .reek.yml file that excludes certain paths
- When I run rubycritic
  - it does not analyse files with reek that are meant to be excluded
  - it does not add smells from reek for files that are meant to be excluded

PR #538, opened 2025-10-23
fix: Respect exclude_paths in reek config

fbuys:
Add ability to exclude files from Reek analysis using the `exclude_paths` option in `.reek.yml`. Updated the analyser to skip excluded files and added tests to verify excluded files are ignored. Also added a sample excluded file for testing.

Fixes: https://github.com/whitesmith/rubycritic/issues/537

Check list:

### DIFF
--- a/test/lib/rubycritic/analysers/smells/reek_test.rb
+++ b/test/lib/rubycritic/analysers/smells/reek_test.rb
@@ -6,9 +6,17 @@ require 'rubycritic/analysers/smells/reek'
 describe RubyCritic::Analyser::ReekSmells do
   context 'when analysing a smelly file' do
     before do
-      pathname = Pathname.new('test/samples/reek/smelly.rb')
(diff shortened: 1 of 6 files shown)


test/lib/rubycritic/analysers/smells/reek_test.rb
```diff
@@ -6,9 +6,17 @@
 describe RubyCritic::Analyser::ReekSmells do
   context 'when analysing a smelly file' do
     before do
-      pathname = Pathname.new('test/samples/reek/smelly.rb')
-      @analysed_module = AnalysedModuleDouble.new(pathname: pathname, smells: [])
-      analysed_modules = [@analysed_module]
+      # Reset Reek configuration between tests
+      # Config is cached and can leak between tests
+      RubyCritic::Reek.instance_variable_set(:@configuration, nil)
```
fbuys: Key to fixing flaky tests

PR #566, opened 2026-01-26
Additional changes for 5.0 (Ruby 4 support) branch

faisal:
This PR contains a number of changes on top of when we last touched the 5.0 branch:

- It updates the version, and it updates the version compatibility matrix in README.md (to match what we already have in the 4.12.0 release)
- It runs mdl in CI using Ruby 4 (which we'd had in the past but dropped in branch shuffles)
- It moves ostruct from the development dependencies to the runtime dependencies because Ruby 4 drops ostruct.
- It updates the byebug dependency to a version with fixes with Ruby 4 support.
- Minor changelog weaks beyond just adding the above

- [ ] [Squash all commits into a single one](https://github.com/whitesmith/rubycritic/blob/main/CONTRIBUTING.md)

This is kind of a mess and we'll need to clean up the final PR, but we had that problem anyway.


etagwerker (approved):
@faisal Looks good, thank you! 💯

PR #570, opened 2026-04-02
Drop Ruby 3.2 support

faisal:
Ruby 3.2 is now EOL. This is a PR to drop support for it.

n.b. I do not see a driver to drop such support right now, but I'm opening it so we've done the work for when it's time.

- Update minimum Ruby requirement to 3.2
- Drop Ruby 3.2 from CI matrix
- Adopt some RubyCritic-provided corrections, but suppress Style/OneClassPerFile in all its flagged files

Check list:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,7 +28,6 @@ jobs:
       fail-fast: false
       matrix:
         ruby-version:
-          - '3.2'
           - '3.3'
           - '3.4'
           - '4.0'
@@ -59,7 +58,6 @@ jobs:
       fail-fast: false
       matrix:
         ruby-version:
-          - '3.2'
           - '3.3'
           - '3.4'
           - '4.0'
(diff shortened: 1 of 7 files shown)


etagwerker (approved):
@faisal Looks good! 

What version do you suggest we should release after this change? 6.0.0 or 5.1.0? 

I'd like it to make it clear that 5.0.0 is the last version to support Ruby 3.2

faisal:
> What version do you suggest we should release after this change? 6.0.0 or 5.1.0?

Given the lack of functional or major structural changes, I think 5.1.0 makes sense.

> I'd like it to make it clear that 5.0.0 is the last version to support Ruby 3.2

I should update the README to note the compatibility change.

JuanVqz:
Hello,

My perspective on this matter is that we should not discontinue the support for EOL Ruby versions. A broader Ruby support base is advantageous as it enables more applications to utilize the gem, specially old apps wanting to get out of tech debt

faisal:
No disagreement here -- I just prepped this so it was ready if we find a reason.

I'll move it to draft for now.

PR #571, opened 2026-05-24
Replace Aruba with direct API calls

faisal:
Aruba is a recurring source of dependency conflicts when updating Rubycritic dependencies*. Update Aruba call sites to call the underlying command line calls directly. Call with Open3 so as to capture results and return status.

`*` Most recently, being able to update to diff-lcs 2.0, which we do in this commit. We ran into it more before when Rubycritic relied on Cucumber, though it doesn't any more.

Check list:

### DIFF
--- a/test/integration/integration_test_helper.rb
+++ b/test/integration/integration_test_helper.rb
@@ -1,21 +1,41 @@
 # frozen_string_literal: true
 
 require_relative '../test_helper'
-require 'aruba/api'
+require 'open3'
+require 'timeout'
+require 'tmpdir'
+require 'fileutils'
 require_relative '../../lib/rubycritic'
 require_relative '../../lib/rubycritic/cli/application'
 require_relative '../../lib/rubycritic/commands/status_reporter'
 
-Aruba.configure { |config| config.exit_timeout = 30 }
-
 module IntegrationTestHelper
-  include Aruba::Api
+  GEM_ROOT = File.expand_path('../..', __dir__)
+  COMMAND_TIMEOUT = 30
+
+  CommandResult = Struct.new(:stdout, :stderr, :exit_status)
+
+  def before_setup
+    super
+    @sandbox_dir = Dir.mktmpdir('rubycritic-spec-')
+    @original_pwd = Dir.pwd
+    Dir.chdir(@sandbox_dir)
+  end
+
+  def after_teardown
+    Dir.chdir(@original_pwd) if @original_pwd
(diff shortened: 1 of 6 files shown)


faisal:
@etagwerker, @JuanVqz Would you rather remove the Aruba requirement, or just manage its progression?

test/integration/integration_test_helper.rb
```diff
@@ -61,4 +78,14 @@ def foo; end
...
   end
+
+  private
+
+  def run_command(command)
+    env = { 'PATH' => "#{File.join(GEM_ROOT, 'bin')}#{File::PATH_SEPARATOR}#{ENV.fetch('PATH', '')}" }
+    stdout, stderr, status = Timeout.timeout(COMMAND_TIMEOUT) do
+      Open3.capture3(env, command)
+    end
+    CommandResult.new(stdout, stderr, status.exitstatus)
```
JuanVqz: One small, non-blocking note:

`Timeout.timeout` wraps `Open3.capture3`, but on a timeout the spawned child process isn't killed, just orphaned. If you want to tighten it up, switching to `Open3.popen3` and `Process.kill`-ing `wait_thr.pid` in a rescue `Timeout::Error` handles that.

Totally optional given how fast these specs run but to match Aruba killing on exit_timeout.

```suggestion
    Open3.popen3(env, command) do |stdin, stdout, stderr, wait_thr|
      stdin.close
      out = err = nil
      begin
        Timeout.timeout(COMMAND_TIMEOUT) do
          out = stdout.read
          err = stderr.read
          wait_thr.join
        end
      rescue Timeout::Error
        Process.kill('KILL', wait_thr.pid)
        wait_thr.join
        raise
      end
      return CommandResult.new(out, err, wait_thr.value.exitstatus)
    end
```

Happy to approve either way.

JuanVqz (commented):
@faisal Thanks for this.

The justification checks out: with the Cucumber features already gone, aruba is the only dev dependency still pulling in `cucumber` and `rspec-expectations`, both of which cap `diff-lcs at < 2.0`.

Removing aruba is what unblocks `diff-lcs ~> 2.0`, and I confirmed `Diff::LCS.sdiff`/`unchanged?` (the only API we use in test_helper.rb) still works on `diff-lcs 2.0` CI is green across all Ruby versions.

To answer your question: let's remove aruba, this PR does it cleanly.

cc @etagwerker

JuanVqz (approved):

faisal:
Took @JuanVqz's tweak (thank you!), suppressed the resulting rubocop warning, and squashed down to one commit.

etagwerker (approved):
@faisal Looks good, thanks!

